PR #276, opened 2023-02-27
Display version information in the demo settings

waliid:
# Pull request

## Description

This PR allows us to display the version and the build number within the demo app settings.

## Changes made

- Display version and build number

## Checklist

- [X] APIs have been properly documented (if relevant).
- [X] The documentation has been updated (if relevant).
- [X] New unit tests have been written (if relevant).
- [X] The demo has been updated (if relevant).
- [X] The playground has been updated (if relevant).

### DIFF
--- a/Demo/Sources/SettingsView.swift
+++ b/Demo/Sources/SettingsView.swift
@@ -26,12 +26,19 @@ struct SettingsView: View {
     @AppStorage(UserDefaults.audiovisualBackgroundPlaybackPolicyKey)
     private var audiovisualBackgroundPlaybackPolicyKey: AVPlayerAudiovisualBackgroundPlaybackPolicy = .automatic
 
+    private var version: String {
+        Bundle.main.infoDictionary!["CFBundleShortVersionString"] as! String
+    }
+
+    private var buildVersion: String {
+        Bundle.main.infoDictionary!["CFBundleVersion"] as! String
+    }
+
     var body: some View {
         List {
             applicationSection()
             playerSection()
             debuggingSection()
-            versionSection()
         }
         .navigationTitle("Settings")
     }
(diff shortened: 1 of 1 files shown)


Demo/Sources/SettingsView.swift
```diff
@@ -87,6 +88,19 @@ struct SettingsView: View {
...
     }
 
+    @ViewBuilder
+    private func versionSection() -> some View {
+        Section {
+            EmptyView()
+        } footer: {
+            HStack {
+                Text("Version \(Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String ?? "")")
+                Text("Build \(Bundle.main.infoDictionary?["CFBundleVersion"] as? String ?? "")")
```
defagos, line 98: Improvement proposals:

1. These values should be available from the Info.plist so force-unwrapping should be safe and crash if we failed to set these keys.
2. Extracting as private vars could make it even more readable.

So for example:

```swift
private var version: String {
    Bundle.main.infoDictionary!["CFBundleShortVersionString"] as! String
}
```

and similarly for the build number.

Demo/Sources/SettingsView.swift
```diff
@@ -87,6 +88,19 @@ struct SettingsView: View {
         }
     }
 
+    @ViewBuilder
+    private func versionSection() -> some View {
```
defagos, line 92: The layout would probably be even nicer if the footer is attached to the last section (currently the debugging one):

```swift
Section {
    // ...
} header: {
    Text("Debugging")
} footer: {
    HStack {
        // ...
    }
    .frame(maxWidth: .infinity)
}
```

defagos (changes requested):
Please see if the attached improvement proposals make sense.

defagos (approved):

PR #341, opened 2023-04-07
Add full screen support to the demo player

defagos:
# Pull request

## Description

This PR adds full screen support to the main demo player. Full-screen mode is currently only enabled for the playlist demo (in other settings the player fills the screen at all times).

## Changes made

- Add full screen button to `PlaybackView`.
- Add playback view layouts (with a special `.inline` layout for integrations without full screen button needs).
- Control layout with a binding.
- Add quasi-invisible background to bottom controls to avoid taps in transparent areas triggering the user interface toggle tap gesture.

## Checklist

### DIFF
--- a/Demo/Sources/Players/PlaybackView.swift
+++ b/Demo/Sources/Players/PlaybackView.swift
@@ -13,6 +13,7 @@ import SwiftUI
 @available(tvOS, unavailable)
 private struct MainView: View {
     @ObservedObject var player: Player
+    @Binding var layout: PlaybackView.Layout
     @StateObject private var visibilityTracker = VisibilityTracker()
 
     @State private var layoutInfo: LayoutInfo = .none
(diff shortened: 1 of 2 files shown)


Demo/Sources/Players/PlaybackView.swift
```diff
@@ -287,7 +327,10 @@ private struct TimeBar: View {
             routePickerView()
             TimeSlider(player: player, progressTracker: progressTracker)
             LiveLabel(player: player, progressTracker: progressTracker)
+            FullScreenButton(layout: $layout)
         }
+        // Prevent taps from going through in transparent areas
+        .background(Color(white: 1, opacity: 0.0001))
```
waliid, line 333:
```suggestion
        .contentShape(Rectangle())
```

defagos: Sadly the behavior is not the same. Taps can still go through with `contentShape(_:)`, which means tapping on a transparent area next to the full screen double arrow will trigger not only the full screen toggle button, but also the user interface visibility gesture, which is awkward.

waliid (approved):

PR #421, opened 2023-06-19
Drive demo navigation with iOS 16 navigation API

defagos:
# Description

This PR revisits app navigation with the new iOS 16 navigation APIs.

# Changes made

- Update navigation links to use destinations.
- Introduce router to manage global routing for each tab.
- Introduce helpers to make routing management simple.
- Revisit cells so that they do not manage the same state but rather provide a way to alter the routes at the parent level.
- Extract types related to content lists.

# Checklist


Demo/Sources/Router/Router.swift
```diff
@@ -0,0 +1,78 @@
...
+/// Manages application routes.
+///
+/// You can manage navigation with a `RoutedNavigationStack` or retrieve the router through the environment
+/// to present or dismiss the associated modal.
+final class Router: ObservableObject {
+    @Published fileprivate var path: [RouterDestination] = []
+    @Published fileprivate var presented: RouterDestination?
+
+    func present(_ destination: RouterDestination) {
+        presented = destination
```
waliid, line 72: By doing that, we get the following run time issue: 

_Publishing changes from within view updates is not allowed, this will cause undefined behavior._

I don't really know if it's a real issue or an Xcode 17 Beta issue.

defagos: Same issue with the iOS 17.0 iPhone 14 simulator, not with the iPhone 14 Pro simulator or on device, or when running with Xcode 14. The code itself is also not evaluated as part of a body, this is why I think this is likely an Xcode 15 beta issue.

waliid (approved):

PR #437, opened 2023-06-30
Add sample streams

defagos:
# Description

This PR adds some [sample streams](https://demo.unified-streaming.com/k8s/features/stable/#!/hls) from Unified Streaming.

Several streams do not play well on mobile and were not included.

# Changes made

Self-explanatory.

# Checklist

### DIFF
--- a/Demo/Sources/Model/Template.swift
+++ b/Demo/Sources/Model/Template.swift
@@ -14,161 +14,159 @@ enum URLTemplate {
     static let onDemandVideoHLS = Template(
         title: "Switzerland says sorry! The fondue invasion",
         description: "VOD - HLS",
-        type: .url(.init(string: "https://swi-vod.akamaized.net/videoJson/47603186/master.m3u8")!)
+        type: .url("https://swi-vod.akamaized.net/videoJson/47603186/master.m3u8")
     )
     static let shortOnDemandVideoHLS = Template(
         title: "Des violents orages ont touché Ajaccio, chef-lieu de la Corse, jeudi",
         description: "VOD - HLS (short)",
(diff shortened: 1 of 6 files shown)


Demo/Sources/Model/Template.swift
```diff
@@ -9,153 +9,167 @@ import CoreBusiness
...
 // Apple streams are found at https://developer.apple.com/streaming/examples/
+// Unified Streaming streams are found at https://demo.unified-streaming.com/k8s/features/stable/#!/hls
 enum URLTemplate {
     static let onDemandVideoHLS = Template(
         title: "Switzerland says sorry! The fondue invasion",
         description: "VOD - HLS",
-        type: .url(
-            URL(string: "https://swi-vod.akamaized.net/videoJson/47603186/master.m3u8")!
-        )
+        type: .url(.init(string: "https://swi-vod.akamaized.net/videoJson/47603186/master.m3u8")!)
```
waliid, line 17: # Suggestion 

What do you think about creating a private extension of `URL` which implement `ExpressibleByStringLiteral` in this file ?

We might have something like that: 

```swift
private extension URL: ExpressibleByStringLiteral {
    public init(stringLiteral value: StringLiteralType) {
        self = URL(stringLiteral: value)!
    }
}
```

In this case, we will be able to call

```swift 
.url("https://swi-vod.akamaized.net/videoJson/47603186/master.m3u8")
```

instead of calling 

```swift
.url(.init(string: "https://swi-vod.akamaized.net/videoJson/47603186/master.m3u8")!)
```

defagos: This makes sense. Maybe we could even update `Media.Type.url` to accept as `String` as parameter instead of a `URL`? In general I am not fond of this approach but in this case this is mostly a matter of providing some static URLs to our player, so I guess this could be acceptable as well.

waliid (approved):

PR #439, opened 2023-07-02
Add loaded time ranges

waliid:
# Description

This PR allows us to know the loaded time ranges by the player.

# Changes made

- The progress tracker can now provide the loaded ratio based on the player's loaded time ranges.
- A new slider (`PlaybackSlider)` has been added in the demo app.

## Old

old-slider

## New

new-slider

# Checklist


Sources/Player/Publishers/AVPlayerItemPublishers.swift
```diff
@@ -36,6 +36,21 @@ extension AVPlayerItem {
...
 
+    func loadedTimeRangePublisher() -> AnyPublisher<CMTimeRange, Never> {
+        Publishers.CombineLatest(
+            publisher(for: \.status),
+            publisher(for: \.loadedTimeRanges)
+        )
+        .map { status, loadedTimeRanges in
+            guard status == .readyToPlay else { return .zero }
+            let start = loadedTimeRanges.first?.timeRangeValue.start ?? .zero
+            let end = loadedTimeRanges.first?.timeRangeValue.end ?? .zero
```
defagos, line 47:
```suggestion
            let end = loadedTimeRanges.last?.timeRangeValue.end ?? .zero
```

Sources/Player/Publishers/AVPlayerItemPublishers.swift
```diff
@@ -36,6 +36,21 @@ extension AVPlayerItem {
         .eraseToAnyPublisher()
     }
 
+    func loadedTimeRangePublisher() -> AnyPublisher<CMTimeRange, Never> {
```
defagos, line 39: From an API consistency perspective with other publishers like `timeRangePublisher()` maybe we should return `.invalid` when the range is unknown?

Sources/Player/Publishers/AVPlayerPublishers.swift
```diff
@@ -59,6 +59,17 @@ extension AVPlayer {
             .eraseToAnyPublisher()
     }
 
+    func currentItemLoadedTimeRangePublisher() -> AnyPublisher<CMTimeRange, Never> {
+        publisher(for: \.currentItem)
+            .map { item in
+                guard let item else { return Just(CMTimeRange.zero).eraseToAnyPublisher() }
```
defagos, line 65: Return `.invalid` when not known?

Sources/Player/UserInterface/ProgressTracker.swift
```diff
@@ -27,6 +27,8 @@ public final class ProgressTracker: ObservableObject {
         }
     }
 
+    @Published public var loaded: Float = .zero
```
defagos, line 30: Could maybe be confused with an loaded state, maybe calling this value `buffer` (or something similar) might be clearer?

Moreover the documentation is missing.

Demo/Sources/Players/PlaybackView.swift
```diff
@@ -435,12 +428,12 @@ private struct TimeSlider: View {
         ._debugBodyCounter(color: .blue)
     }
 
-    private static func formattedDuration(_ duration: TimeInterval) -> String {
-        if duration < 60 * 60 {
-            return shortFormatter.string(from: duration)!
+    private static func formattedDuration(_ time: TimeInterval, totalDuration: TimeInterval) -> String {
```
defagos, line 431: Smart, this is intended to avoid frame changes when changing the progress I guess.

Since there is now a duration parameter we could maybe update the method prototype, something like `formattedTime(_ time: TimeInterval, duration: TimeInterval)`?

Demo/Sources/Views/PlaybackSlider.swift
```diff
@@ -0,0 +1,118 @@
...
+        }
+        .frame(maxWidth: .infinity, minHeight: 10, maxHeight: 10)
+    }
+
+    init(progressTracker: ProgressTracker, minimumValueText: String? = nil, maximumValueText: String? = nil) {
+        self.progressTracker = progressTracker
+        self.minimumValueText = minimumValueText
+        self.maximumValueText = maximumValueText
+        _value = Binding(progressTracker, at: \.progress)
+        _buffer = Binding(progressTracker, at: \.loaded)
```
defagos, line 43: The `buffer` rename proposed elsewhere seems to fit here nicely indeed.

Demo/Sources/Views/PlaybackSlider.swift
```diff
@@ -0,0 +1,118 @@
...
+        let widthToApply = CGFloat(buffer) * width
+        if widthToApply.isNormal {
+            bufferWidth = widthToApply
+        }
+    }
+}
+
+struct PlaybackSlider_Previews: PreviewProvider {
+    static var previews: some View {
+        PlaybackSlider(progressTracker: .init(interval: .zero))
```
defagos, line 114: The preview background should be forced to black (or the preview itself to dark mode) so that the slider is visible.

Demo/Sources/Views/PlaybackSlider.swift
```diff
@@ -0,0 +1,118 @@
...
+    var body: some View {
+        HStack {
+            text(minimumValueText)
+            progressBar(valueWidth: valueWidth, bufferWidth: bufferWidth)
+            text(maximumValueText)
+        }
+        .frame(maxWidth: .infinity, minHeight: 10, maxHeight: 10)
+    }
+
+    init(progressTracker: ProgressTracker, minimumValueText: String? = nil, maximumValueText: String? = nil) {
```
defagos, line 38: `minimumValueLabel` and `maximumValueLabel` parameters would provide API consistency with the native `Slider`, which would be quite nice IMHO. `EmptyView` specializations could then be introduced to implement optionality.

Demo/Sources/Views/PlaybackSlider.swift
```diff
@@ -0,0 +1,118 @@
...
+    private func text(_ text: String?) -> some View {
+        if let text {
+            Text(text)
+                .font(.caption)
+                .monospacedDigit()
+        }
+    }
+
+    @ViewBuilder
+    private func rectangle(opacity: Double = 1, width: CGFloat? = nil) -> some View {
```
defagos, line 56:
```suggestion
    private func rectangle(opacity: Double, width: CGFloat? = nil) -> some View {
```

Demo/Sources/Views/PlaybackSlider.swift
```diff
@@ -0,0 +1,118 @@
...
+    @State private var previousValueWidth: CGFloat = 0
+
+    @Binding private var buffer: Float
+    @State private var bufferWidth: CGFloat = 0
+
+    @State private var isDragging = false {
+        didSet {
+            progressTracker.isInteracting = isDragging
+        }
+    }
```
defagos, line 27: Probably superfluous, we can probably mutate `progressTracker.isInteracting` directly. The `progressTracker` should likely be made an `@ObservedObject` in the process.

Demo/Sources/Views/PlaybackSlider.swift
```diff
@@ -0,0 +1,118 @@
...
+        }
+        .cornerRadius(5)
+        .scaleEffect(y: isDragging ? 1.5 : 1.0)
+        .animation(.smooth(duration: 0.4), value: isDragging)
+    }
+
+    private func dragGesture(geometry: GeometryProxy) -> some Gesture {
+        DragGesture()
+            .onChanged { gesture in
+                isDragging = true
```
defagos, line 83: Maybe it would be even possible to enter interaction mode on touch down (similar to the native player UI experience)?

Demo/Sources/Views/PlaybackSlider.swift
```diff
@@ -0,0 +1,118 @@
...
+
+#if os(iOS)
+struct PlaybackSlider: View {
+    let progressTracker: ProgressTracker
+    let minimumValueText: String?
+    let maximumValueText: String?
+
+    @Binding private var value: Float
+    @State private var valueWidth: CGFloat = 0
+    @State private var previousValueWidth: CGFloat = 0
```
defagos, line 18: Could we maybe get rid of this state? Not sure why this would be necessary if we can access the drag gesture start location.

waliid: If we want the same behavior as the Apple Slider, we should keep the previous value. 
This will prevent jumping directly to the starting gesture location.

Demo/Sources/Views/PlaybackSlider.swift
waliid: Yep! However, it seems to work without putting the `progressTracker` as `@ObservedObject`.

Demo/Sources/Views/PlaybackSlider.swift
waliid: It's a good idea! What do you think about creating an issue for that? 
It may not be easy to handle it very quickly.

defagos: Sure, the current result is great already.

defagos (approved):

defagos (approved):

defagos (approved):

defagos (approved):

defagos (approved):

Issue #473, opened 2023-07-21
Seeks gets stuck with some content


defagos:
### Description of the problem

When playing a TV series episode from RSI or a documentary fro RTS, seeking does not work anymore with the Pillarbox player and our custom UI. The player gets stuck.

### Reproducibility

Always

### Steps to reproduce

1. Open Pillarbox demo.
2. Check that the custom player and immediate seeking are enabled in the app settings.
3. Open a TV series episode from RSI or an RTS documentary.
4. Seek anywhere in the content.

If you repeat the process with the system UI or with deferred seek behavior the problem does not arise.

### Library version

Build 292

### Operating system

iOS 16

### Is there an existing issue for this?

- [X] I have searched existing issues and found no similar bug report.

defagos:
The issue is related to recent seek improvements (#466). The code identifies that the content is able to play forward, which is confirmed by the presence of an I-frame playlist:

```
#EXTM3U

#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="audio0",NAME="Italiano",LANGUAGE="it",AUTOSELECT=YES,DEFAULT=YES,CHANNELS="2",URI="index-f1-a1.m3u8?aka_me_session_id=AAAAAAAAAAB5E7xkAAAAAPIMy8k5gckTPSs9YtwLv1M5A3Up8%2ffZBLyEW+TsFdyR9DTLh0vyXgBVWjH8EO0BB8NutCV1hpHt&aka_media_format_type=hls"


#EXT-X-MEDIA:TYPE=SUBTITLES,GROUP-ID="subs0",NAME="Italiano",LANGUAGE="it",AUTOSELECT=YES,DEFAULT=YES,URI="index-f3.m3u8?aka_me_session_id=AAAAAAAAAAB5E7xkAAAAAPIMy8k5gckTPSs9YtwLv1M5A3Up8%2ffZBLyEW+TsFdyR9DTLh0vyXgBVWjH8EO0BB8NutCV1hpHt&aka_media_format_type=hls"

#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=1359955,RESOLUTION=640x360,FRAME-RATE=24.972,CODECS="avc1.640028,mp4a.40.2",VIDEO-RANGE=SDR,AUDIO="audio0",SUBTITLES="subs0"
[...]

PR #476, opened 2023-07-24
Further improve seek user experience

defagos:
# Description

This PR further improves the seek user experience to have it on par with the one provided by the system player.

If running the code in its prior state one could namely observe that the experience was still lagging behind the standard system player user experience when seeking while playing (images were not moving as smoothly, e.g. with MP4 step-by-step seeking). But if playback was paused before seeking then both our player and the system player were providing the same experience.

The reason was that, even though we were pausing / resuming for each global seek group (current seek and pending seeks), we weren't pausing during the whole user interaction. Since an interaction leads to several seek groups being created, when each group completed playback was resumed, leading to a less smooth experience than if playback was kept paused the whole time.

This change also coincidentally mitigates issues reported in #473, though stream issues still have to be fixed.

# Changes made

- Implement pause during seeks in `ProgressTracker` to associate it with user interaction.
- Remove pause from `QueuePlayer` seeks.

# Checklist

### DIFF
--- a/Sources/Player/UserInterface/ProgressTracker.swift
+++ b/Sources/Player/UserInterface/ProgressTracker.swift
@@ -18,16 +18,24 @@ public final class ProgressTracker: ObservableObject {
     /// The player to attach.
     ///
     /// Use `View.bind(_:to:)` in SwiftUI code.
-    @Published public var player: Player?
+    @Published public var player: Player? {
+        willSet {
+            resumePlaybackIfNeeded(with: player)
+        }
+        didSet {
+            guard isInteracting else { return }
+            pausePlaybackIfNeeded(with: player)
+        }
+    }
 
     /// A Boolean describing whether user interaction is currently changing the progress value.
     @Published public var isInteracting = false {
         willSet {
             if newValue {
-                pausePlaybackIfNeeded()
+                pausePlaybackIfNeeded(with: player)
             }
             else {
-                resumePlaybackIfNeeded()
+                resumePlaybackIfNeeded(with: player)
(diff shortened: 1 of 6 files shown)


Sources/Player/UserInterface/ProgressTracker.swift
```diff
@@ -138,6 +152,18 @@ public final class ProgressTracker: ObservableObject {
         guard let timeRange, let progress else { return nil }
         return timeRange.start + CMTimeMultiplyByFloat64(timeRange.duration, multiplier: Float64(progress))
     }
+
+    private func pausePlaybackIfNeeded(with player: Player?) {
+        guard let player, player.playbackState == .playing else { return }
+        player.pause()
+        wasPaused = true
```
waliid, line 159:
```suggestion
        wasPaused.toggle()
```

Sources/Player/UserInterface/ProgressTracker.swift
```diff
@@ -138,6 +152,18 @@ public final class ProgressTracker: ObservableObject {
...
+    }
+
+    private func resumePlaybackIfNeeded(with player: Player?) {
+        guard let player, wasPaused else { return }
+        player.play()
+        wasPaused = false
```
waliid, line 165:
```suggestion
        wasPaused.toggle()
```

waliid (approved):

Sources/Player/UserInterface/ProgressTracker.swift
defagos: Probably a bit safer to force the value.

PR #480, opened 2023-07-25
Add YouTube auto paste

waliid:
# Description

This PR allows to auto paste YouTube URLs in the `TextField`, if any detected.

# Changes made

- A `youTubePaste` method has been added as `View` extension.

# Improvement ideas

- Use `UIPasteControl` to avoid raising an alert.
- Try to use continuity to copy/paste from Apple devices.

# Checklist

### DIFF
--- a/Demo/Sources/YouTube/YouTube+View.swift
+++ b/Demo/Sources/YouTube/YouTube+View.swift
@@ -8,11 +8,18 @@ import SwiftUI
 import YouTubeIdentifier
 
 extension View {
+    private static func paste(_ text: Binding<String>) {
+        guard let paste = UIPasteboard.general.string,
+              YouTubeIdentifier.extract(from: URL(string: paste)!) != nil else { return }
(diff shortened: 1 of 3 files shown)


Demo/Sources/YouTube/YouTube+View.swift
```diff
@@ -0,0 +1,40 @@
...
+        self
+#endif
+    }
+}
+
+#if os(iOS)
+private extension UIPasteboard {
+    var hasYouTubeUrl: Bool {
+        guard let string else { return false }
+        return YouTubeIdentifier.extract(from: URL(string: string)!) != nil
```
defagos, line 37:
```suggestion
        guard let string, let url = URL(string: string) else { return false }
        return YouTubeIdentifier.extract(from: url) != nil
```

is probably safer.

defagos (approved):

defagos (approved):

PR #594, opened 2023-10-09
Fix flaky tests

defagos:
# Description

This PR makes some failing tests more reliable.

# Changes made

Here are some issues I mitigated:

- It seems that test suites run in parallel might concurrently alter `MediaAccessibility` settings, leading to unreliable results. For this reason I ensured that only one test suite messes with `MediaAccessibility` settings by having a test for media selection preference reset only check the audible case, which should be sufficient.
- I upgraded the error test by waiting until an error has been received. A timeout of 2 seconds seems too short.
- I removed less relevant expectations from _skip to default_ tests, either redundant with the _can skip_ tests, or not really interesting (time for a livestream without DVR).
- I updated the number of retries to 3, in an attempt to mitigate those test which are a bit flaky on the CI but not on our development machines. This value matches the standard number of retries for test plans, which we cannot conveniently use with packages AFAIK.

The comScore test suite is definitely a bit more flaky, maybe because results are subject to network conditions (events are not emitted if no network is available).

I also checked whether test parallelization could possibly make comScore tests more flaky but found no proof of this conjecture.

# Checklist

### DIFF
--- a/.swiftpm/xcode/xcshareddata/xcschemes/Pillarbox-Package-NoDebug.xcscheme
+++ b/.swiftpm/xcode/xcshareddata/xcschemes/Pillarbox-Package-NoDebug.xcscheme
@@ -159,6 +159,7 @@
       <Testables>
          <TestableReference
             skipped = "NO"
+            parallelizable = "YES"
             testExecutionOrdering = "random">
             <BuildableReference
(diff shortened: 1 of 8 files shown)


waliid (approved):
🙏

PR #604, opened 2023-10-16
Display the URL cache size

waliid:
# Description

Self-explanatory.

clear_cache

# Changes made

- Added `UrlCacheView`.
- Updated `SettingView`.

# Checklist

### DIFF
--- a/Demo/Sources/Settings/SettingsView.swift
+++ b/Demo/Sources/Settings/SettingsView.swift
@@ -8,6 +8,31 @@ import AVFoundation
 import Player
 import SwiftUI
 
+private struct UrlCacheView: View {
+    @State private var urlCacheSize: String = ""
+
+    var body: some View {
+        HStack {
+            Button("Clear URL cache", action: clearUrlCache)
+            Spacer()
+            Text(urlCacheSize)
+                .font(.footnote)
+        }
+        .onAppear {
+            updateCacheSize()
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


Demo/Sources/Settings/SettingsView.swift
```diff
@@ -8,6 +8,35 @@ import AVFoundation
...
+                .buttonStyle(.borderless)
+#endif
+            Spacer()
+            Text(urlCacheSize)
+                .font(.footnote)
+                .foregroundColor(.red)
+        }
+        .onAppear {
+            updateCacheSize()
+        }
```
defagos, line 27:
```suggestion
        .onAppear(perform: updateCacheSize)
```

defagos, line 27: The `currentDiskUsage` property seems to be KVObservable. You can therefore likely use `.onReceive(URLCache.shared.publisher(for: \.currentDiskUsage))` to not only get the initial value, but also potential updates which could happen while the view is visible on screen.

waliid: Thank you for your suggestion. It was the first thing I attempted. Unfortunately, it didn't work. I'll give it another try and inform you of the results.

defagos: You're right. The current implementation is good enough, then, thanks.

defagos (approved):

PR #631, opened 2023-11-15
Support Picture in Picture in playlists 

waliid:
# Description

Self-explanatory.

# Changes made

- The view model has been marked as an `ObservedObject`
- The view model is now instantiated via a shared instance.

# Result

https://github.com/SRGSSR/pillarbox-apple/assets/3347810/839f3c9a-9049-4091-9137-48061ade49e4

# Checklist

### DIFF
--- a/Demo/Sources/Router/Router.swift
+++ b/Demo/Sources/Router/Router.swift
@@ -52,10 +52,10 @@ final class Router: ObservableObject {
 extension Router: PictureInPictureDelegate {
     func pictureInPictureWillStart() {
         switch presented {
-        case .player, .systemPlayer:
+        case .player, .systemPlayer, .playlist:
             previousPresented = presented
             presented = nil
-        case .inlineSystemPlayer, .playlist:
+        case .inlineSystemPlayer:
             previousPresented = presented
         default:
             break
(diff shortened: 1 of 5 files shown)


Demo/Sources/Router/Router.swift
defagos, line 1: I would have expected auto-dismissal for the custom layout, but I guess the choice has been made so that the behavior works for the system player as well.

Proposal: To keep things simple we could support only one layout for the playlist, maybe our custom one. This would require being able to provide the `playerLayout` as parameter of `PlaybackView`. The in-app layout setting would then only be read in `PlayerView`.

Demo/Sources/Showcase/Playlist/PlaylistView.swift
defagos, line 1: I could experience the issue you reported. Removing the `contentShape` modifier we can still fix the issue directly in our library, as follows:

```swift
extension View {
    func onWillAppear(perform action: @escaping () -> Void) -> some View {
        background {
            WillAppearView(action: action)
                .allowsHitTesting(false)
        }
    }
}
```

MGaetan89 (approved):

defagos (approved):

PR #636, opened 2023-11-17
Fix player configuration update in Picture-in-Picture implementations

waliid:
# Description

This PR addresses the issue related to the Picture in Picture implementation. 
Currently, the view model is shared, so the initialization occurs only once. 
Consequently, the player configuration is also done only once, leading to a situation where changes made on the settings page (smart navigation, allows external playback, etc.) have no impact on the player.

# Changes made

- ~Created a new player when items changed.~
- A new method `applyConfiguration` has been introduced to facilitate the player configuration update.

# Checklist

### DIFF
--- a/Demo/Sources/Players/PlayerViewModel.swift
+++ b/Demo/Sources/Players/PlayerViewModel.swift
@@ -11,7 +11,7 @@ final class PlayerViewModel {
         didSet {
             guard media != oldValue else { return }
             if let playerItem = media?.playerItem() {
-                player = Player(items: [playerItem], configuration: .standard)
+                player.items = [playerItem]
             }
             else {
                 player.removeAllItems()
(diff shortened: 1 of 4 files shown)


waliid:
Unfortunately, this approach doesn't work very well!

defagos:
As discussed today not sure we should let the configuration be mutable since this might add some complexity later (e.g. internal updates required by a setting change which are currently avoided since the configuration is never changed).

We can likely improve our configuration, though:

- In the player API we can introduce mutable properties for settings which have no update cost, instead of having them fixed in the configuration.
- In our demo we can probably make things a bit less globally configurable but without impacting our ability to test player features.

This is work we should likely perform on a separate dedicated PR I guess.

defagos:
Closed since #638 merged instead.

PR #644, opened 2023-11-21
Display SRG SSR media duration in minutes

defagos:
# Description

This PR simply adds duration information to SRG SSR content lists, as was done on in demos for other platforms.

# Changes made

Self-explanatory.

# Checklist

### DIFF
--- a/Demo/Sources/Model/MediaDescription.swift
+++ b/Demo/Sources/Model/MediaDescription.swift
@@ -13,12 +13,20 @@ enum MediaDescription {
     }
 
     private static var dateFormatter: DateFormatter = {
-        let dateFormatter = DateFormatter()
-        dateFormatter.timeZone = TimeZone(identifier: "Europe/Zurich")
-        dateFormatter.dateStyle = .long
-        dateFormatter.timeStyle = .none
-        dateFormatter.doesRelativeDateFormatting = true
-        return dateFormatter
+        let formatter = DateFormatter()
+        formatter.timeZone = TimeZone(identifier: "Europe/Zurich")
+        formatter.dateStyle = .long
+        formatter.timeStyle = .none
+        formatter.doesRelativeDateFormatting = true
+        return formatter
+    }()
+
+    private static var minuteFormatter: DateComponentsFormatter = {
+        let formatter = DateComponentsFormatter()
+        formatter.allowedUnits = .minute
+        formatter.unitsStyle = .short
+        formatter.zeroFormattingBehavior = .pad
+        return formatter
     }()
 
     static func title(for media: SRGMedia) -> String {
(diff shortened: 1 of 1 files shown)


Demo/Sources/Model/MediaDescription.swift
```diff
@@ -39,4 +46,16 @@ enum MediaDescription {
     static func style(for media: SRGMedia) -> Style {
         media.timeAvailability(at: Date()) == .available ? .standard : .disabled
     }
+
+    private static func date(for media: SRGMedia) -> String {
+        dateFormatter.string(from: media.date)
+    }
+
+    private static func duration(for media: SRGMedia) -> String {
+        minuteFormatter.string(from: max(media.duration / 1000, 60))!
```
waliid, line 55: I understand this fallback to prevent displaying something like **0 min**.
In this context, is it the application's role to hide invalid or incorrect data? 🤷‍♂️

waliid (approved):

Demo/Sources/Model/MediaDescription.swift
```diff
@@ -13,21 +13,28 @@ enum MediaDescription {
...
+        formatter.timeStyle = .none
+        formatter.doesRelativeDateFormatting = true
+        return formatter
+    }()
+
+    private static var minuteFormatter: DateComponentsFormatter = {
+        let formatter = DateComponentsFormatter()
+        formatter.allowedUnits = .minute
+        formatter.unitsStyle = .short
+        formatter.zeroFormattingBehavior = .pad
```
MGaetan89, line 28: If you go in "Lists" > "TV Topics" > "RTS" > "Séries", you can have durations bigger than 60 min. For example, "L'Adieu" is displayed as "94 min" long.
Wouldn't it be more user friendly to display it as "1 hr, 34 min"?
```suggestion
        formatter.allowedUnits = [.minute, .hour]
        formatter.unitsStyle = .short
        formatter.zeroFormattingBehavior = [.pad, .dropAll]
```

MGaetan89 (approved):

Demo/Sources/Model/MediaDescription.swift
defagos: This is not really invalid, this is the convention we decided to apply in the Play app (items less than 1 min are displayed as _1 min_). I just reused the same idea :)

Demo/Sources/Model/MediaDescription.swift
defagos: Same comment as above. We had the behavior you mention in Play a few years ago but we moved to minute display only since we don't have streams lasting more than a couple of hours. This rule was borrowed as is here.

PR #694, opened 2023-12-12
Improve tvOS navigation

waliid:
# Description

This PR enhances navigation in the fifth main tab of the demo application.

# Changes made

- Replaced `List`s SwiftUI component with `ScrollView`s.
- Added horizontal margins to prevent cell clipping.
- The `navigationTitle` modifier is now exclusively used on iOS.

# Result

| Before | After |
|--------|--------|
| 1-examples-before | 1-examples-after |
|  2-showcase-before | 2-showcase-after |
| 3-lists-before | 3-lists-after |
|  4-search-before | 4-search-after |
|  5-settings-before | 5-settings-after |

# Checklist

### DIFF
--- a/Demo/Sources/Search/SearchView.swift
+++ b/Demo/Sources/Search/SearchView.swift
@@ -40,39 +40,26 @@ struct SearchView: View {
     @ViewBuilder
     private func loadedView(_ medias: [SRGMedia]) -> some View {
         if !medias.isEmpty {
-#if os(iOS)
-            List(medias, id: \.urn) { media in
-                let title = MediaDescription.title(for: media)
-                Cell(title: title, subtitle: MediaDescription.subtitle(for: media), style: MediaDescription.style(for: media)) {
-                    let media = Media(title: media.title, type: .urn(media.urn))
-                    router.presented = .player(media: media)
-                }
-                .onAppear {
-                    if let index = medias.firstIndex(of: media), medias.count - index < kPageSize {
-                        model.loadMore()
-                    }
-                }
-                .swipeActions { CopyButton(text: media.urn) }
-                .scrollDismissesKeyboard(.immediately)
-                .refreshable { await model.refresh() }
-            }
-#else
-            ScrollView {
-                ForEach(medias, id: \.urn) { media in
+            CustomList(data: medias) { media in
+                if let media {
                     let title = MediaDescription.title(for: media)
                     Cell(title: title, subtitle: MediaDescription.subtitle(for: media), style: MediaDescription.style(for: media)) {
                         let media = Media(title: media.title, type: .urn(media.urn))
                         router.presented = .player(media: media)
                     }
-                    .padding(.horizontal, 50)
                     .onAppear {
                         if let index = medias.firstIndex(of: media), medias.count - index < kPageSize {
                             model.loadMore()
                         }
                     }
+                    .padding(.horizontal, constant(iOS: 0, tvOS: 50))
+#if os(iOS)
+                    .swipeActions { CopyButton(text: media.urn) }
+                    .refreshable { await model.refresh() }
+#endif
                 }
             }
-#endif
+            .scrollDismissesKeyboard(.immediately)
         }
         else {
             RefreshableMessageView(model: model, message: "No results.", icon: .empty)


Demo/Sources/ContentLists/ContentListView.swift
```diff
@@ -58,12 +72,17 @@ private struct ContentCell: View {
...
         case let .topic(topic):
+#if os(iOS)
             NavigationLink(
                 topic.title,
                 destination: .contentList(configuration: .init(list: .latestMediasForTopic(topic), vendor: topic.vendor))
             )
-#if os(iOS)
             .swipeActions { CopyButton(text: topic.urn) }
+#else
+            NavigationLink(destination: RouterDestination.contentList(configuration: .init(list: .latestMediasForTopic(topic), vendor: topic.vendor)).view()) {
```
MGaetan89, line 82:
```suggestion
            NavigationLink(destination: .contentList(configuration: .init(list: .latestMediasForTopic(topic), vendor: topic.vendor)).view()) {
```

Demo/Sources/ContentLists/ContentListsView.swift
```diff
@@ -46,7 +59,14 @@ struct ContentListsView: View {
     private static func section(title: String, image: String? = nil, configurations: [ContentList.Configuration]) -> some View {
         Section {
             ForEach(configurations) { configuration in
+#if os(iOS)
                 NavigationLink(configuration.name, destination: .contentList(configuration: configuration))
+#else
+                NavigationLink(destination: RouterDestination.contentList(configuration: configuration).view()) {
```
MGaetan89, line 65:
```suggestion
                NavigationLink(destination: .contentList(configuration: configuration).view()) {
```

Demo/Sources/Examples/ExamplesView.swift
```diff
@@ -80,15 +80,31 @@ struct ExamplesView: View {
...
             MediaEntryView()
             srgSections()
             thirdPartySections()
             miscellaneousSections()
         }
         .scrollDismissesKeyboard(.immediately)
         .animation(.defaultLinear, value: model.protectedMedias)
-        .navigationTitle("Examples")
         .tracked(name: "examples")
         .refreshable { await model.refresh() }
```
MGaetan89, line 109: In `ContentListView`, `refreshable` is applied on iOS only. Is it applied to every platform on purpose here?

defagos:
Two quick feedbacks:

- I think there should be a way to use `List` correctly somehow. I cannot imagine Apple not being able to deliver a cross-platform SwiftUI framework which omits tvOS, so we must be doing something wrong.
- If this is really the case we should avoid littering the code with preprocessor directives testing the platform. Instead I think we should likely create our own abstraction of a `CustomList` which internally uses `List` directly on iOS or a custom implementation on tvOS. This way client code would only need to be updated from `List` to `CustomList`.

Demo/Sources/Search/SearchView.swift
```diff
@@ -51,12 +52,27 @@ struct SearchView: View {
...
+                    Cell(title: title, subtitle: MediaDescription.subtitle(for: media), style: MediaDescription.style(for: media)) {
+                        let media = Media(title: media.title, type: .urn(media.urn))
+                        router.presented = .player(media: media)
+                    }
+                    .padding(.horizontal, 50)
+                    .onAppear {
+                        if let index = medias.firstIndex(of: media), medias.count - index < kPageSize {
+                            model.loadMore()
+                        }
+                    }
```
MGaetan89, line 72: How about creating a dedicated method for that?
```swift
@ViewBuilder
private func content(_ media: SRGMedia, _ medias: [SRGMedia]) -> some View {
    let title = MediaDescription.title(for: media)
    Cell(title: title, subtitle: MediaDescription.subtitle(for: media), style: MediaDescription.style(for: media)) {
        let media = Media(title: media.title, type: .urn(media.urn))
        router.presented = .player(media: media)
    }
#if !os(iOS)
    .padding(.horizontal, 50)
#endif
    .onAppear {
        if let index = medias.firstIndex(of: media), medias.count - index < kPageSize {
            model.loadMore()
        }
    }
#if os(iOS)
    .swipeActions { CopyButton(text: media.urn) }
    .scrollDismissesKeyboard(.immediately)
    .refreshable { await model.refresh() }
#endif
}
```

And the `!media.isEmpty` block of `loadedView` becomes:
```swift
#if os(iOS)
List(medias, id: \.urn) { media in
    content(media, medias)
}
#else
ScrollView {
    ForEach(medias, id: \.urn) { media in
        content(media, medias)
    }
}
#endif
```

defagos:
Might also be handled at the cell level, e.g.:

```swift
import SwiftUI

struct ContentView: View {
    var body: some View {
        List {
            Cell(title: "A")
            Cell(title: "B")
            Cell(title: "C")
            Cell(title: "D")
            Cell(title: "E")
            Cell(title: "F")
            Cell(title: "G")
            Cell(title: "H")
            Cell(title: "I")
            Cell(title: "J")
            Cell(title: "K")
        }
    }
}

struct Cell: View {
    let title: String

    var body: some View {
#if os(tvOS)
        Button(action: {}) {
            Text(title)
                .padding()
                .frame(maxWidth: .infinity, alignment: .leading)
        }
        .buttonStyle(.card)
#else
        Text(title)
#endif
    }
}

#Preview {
    ContentView()
}
```

Here cell highlighting does not adjust colors as in the Settings app so there is still something lacking, but if I remember correctly this is more or less what is being done for managing settings views in the Play app nowadays.

Demo/Sources/ContentLists/ContentListView.swift
waliid: Unfortunately, in this case, the type cannot be inferred because the `NavigationLink` is waiting for a `View`.

Demo/Sources/ContentLists/ContentListsView.swift
waliid: Same as above ☝️

Demo/Sources/Examples/ExamplesView.swift
waliid: It's an oversight on my part, thanks!

defagos:
Using a `List` also seems to ensure more consistent section spacing.

I also tested the demo. The behavior obtained by removing the navigation titles is definitely better. In the settings tab I think that buttons should still take the whole screen width with leading alignment, though.

I recommend you take the Settings app as goal. Even though some choices might differ (e.g. having uppercased section titles) the look & feel for a raw list should basically feel the same, from behavior, to colors or to paddings.

IMHO the formalism to get a list layout on iOS and tvOS should also be identical, hiding differences between custom types (e.g. `Cell` or `CustomList` like described above), with specific differences locally handled with the preprocessor or the trick we use in Play SRG:

```swift
func constant<T>(iOS: T, tvOS: T) -> T {
#if os(tvOS)
    return tvOS
#else
    return iOS
#endif
}
```

I'll stop there, thanks for improving the look & feel of our tvOS demo!

waliid:
I am also surprised that it doesn't work. 
However, it's not so much the lists that don't work, but rather the lists combined with our structure.

Here is our structure:

```swift
TabView {
    NavigationStack {
        List {
        }
    }
}
```

I found a [solution](https://github.com/SRGSSR/pillarbox-apple/commit/0fe4ddd4a1e9a2c833c90190b1bb863e3f023225) to make it working by wrapping everything in a `NavigationStack`.

```swift
NavigationStack {
    TabView {
        NavigationStack {
            List {
            }
        }
    }
}
```

But, it won't make sense anymore when we use a design closer to Android (Card style), as we will be forced to use either a `Grid` or `HStack`.

On tvOS, we will have something like this:

```swift
NavigationStack { // No longer needed with 1. or 2.
    TabView {
        NavigationStack {
            // 1
            ScrollView {
                HStack {
                }
            }
            // OR
            // 2
            ScrollView {
                Grid {
                }
            }
        }
    }
}
```

As the root NavigationStack will no longer be needed, I've supposed that my fix was not really relevant. That's why I started to replace `List`s with `ScrollView`s and `HStack`s to prepare for the next step.

Demo/Sources/Search/SearchView.swift
waliid: Done!

MGaetan89 (approved):

defagos:
The PR has been merged but I am not sure there will be a next step with SwiftUI, except if performance issues with nested views and grid-like structures have been fixed.

IMHO it would probably be easier to do what is required to address the current need, as the future one is likely to require a significant rewrite (and I am not sure a demo justifies this amount of work).

PR #695, opened 2023-12-13
Update cells appearance for tvOS

waliid:
# Description

This PR enables us to update the appearance of cells on tvOS to align with the more common style found on the tvOS platform.

# Changes made

Self-explanatory

| Before | After |
|--------|--------|
|  1-examples-after | examples | 

> Note: The cell's aspects (font, alignments, images, spacing, etc.) will be enhanced in a future PR.

# Checklist

### DIFF
--- a/Demo/Pillarbox-demo.xcodeproj/project.pbxproj
+++ b/Demo/Pillarbox-demo.xcodeproj/project.pbxproj
@@ -13,6 +13,7 @@
 		0ECC5AD52A517A4C0064E701 /* PlaybackSlider.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0ECC5AD42A517A4C0064E701 /* PlaybackSlider.swift */; };
 		0EDFF0DB2B0740DA005030B4 /* SourceCodeViewable.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0EDFF0DA2B0740DA005030B4 /* SourceCodeViewable.swift */; };
 		0EE2A3AE2B29D6D000BAAD65 /* CustomList.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0EE2A3AD2B29D6D000BAAD65 /* CustomList.swift */; };
+		0EE2A3B02B29F82200BAAD65 /* CustomSection.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0EE2A3AF2B29F82200BAAD65 /* CustomSection.swift */; };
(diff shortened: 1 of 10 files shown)


MGaetan89 (approved):

defagos:
Sorry to provide you with my input (I really shouldn't since I am away), but since we already ran into several issues when porting Play SRG from iOS to tvOS I have a few insights:

- Performance with SwiftUI grid has been historically really poor on tvOS. If this is still the case nowadays you will likely not be able to extend the approach applied to the Examples tab in this PR to all screens in the demo app. Check [part 1 of my dedicated article](https://defagos.github.io/swiftui_collection_part1/) for a simple performance test to do on a real tvOS 16+ device. Note that parts 2 and 3 of the article are fun to read but mostly outdated now (there are better ways).
- Focus management is tricky. You cannot navigate vertically between horizontal rows if there is no view to reach. We had to do several tricks in play to redirect the focus in empty spaces using focus guides. There is no such issue with bare vertical lists.
- The grid layout would also totally make sense on a tablet. Usually layout decisions are better made based on user interface size class considerations, not using preprocessor directives (which tend to make the code unnecessarily hard to read).

I don't want to stop the party here but, as suggested by my comment in #694, I strongly think we should keep the demo code simple. We are building an SDK, not the successor of Play SRG. If there is a quick and elegant way to have beautiful grids in the demo that's a no brainer, otherwise I would strongly suggest to take a step back and think about the maintenance costs involved in making the tvOS demo as beautiful as possible **today** (especially if a magic collection API is introduced in later iOS / tvOS versions).

waliid:
Thank you for your feedback.

I've made some tests and I didn't notice any performance issues. 
Have you had a chance to test the code related to this PR? 
I understand that managing focus can be challenging. The question is: is it necessary to address it in our demo app? 

I appreciate your advice and suggestions, but this is an opportunity to learn and explore new things, and also staying aligned with Android TV. 

The code doesn't seem overly complex to maintain, and we can rollback at any time. 
What are your thoughts on this?
Once again, could you please try the demo app with the lazy grid?

defagos:
To avoid losing info since we discussed this matter this morning via Discord:

- My comment was just a warning based on experience, not a call for a rollback.
- If we want to address other platforms like macOS, visionOS or watchOS in the future, it is better if we can stick to a single SwiftUI codebase with minor adjustments per platform, especially for a demo app.

Here is also the sample project I used to report the issue to Apple. I sadly don't have any physical Apple TV device available but the issue seems to affect the tvOS 17.2 simulator running on an M1:

[Stack Grid Performance.zip](https://github.com/SRGSSR/pillarbox-apple/files/13673579/Stack.Grid.Performance.zip)

Issue #696, opened 2023-12-14
Picture in Picture is killed too early if no `VideoView` is initially present


defagos:
### Description of the problem

The current reference counting strategy implemented for PiP sadly has a flaw. We did not notice it because of our demo implementation but in general it will fail if the player view enabled for PiP is initially displayed without `VideoView` actually in the hierarchy.

In this case the reference counter is not increased initially on view entry (since `playerLayer` is `nil`) but will be on exit, leading to early PiP controller deallocation.

### Reproducibility

Always

### Steps to reproduce

1. Edit the `PlaybackView` code so that no `VideoView` is displayed when the stream type is unknown (always the case initially in our demo):

    ```swift
    @ViewBuilder
    private func video() -> some View {
        switch player.mediaType {
        case .audio:
            image(name: "music.note.tv.fill")
        case .video:
            if player.isExternalPlaybackActive {
                image(name: "tv")
            }
            else {
                VideoView(player: player, gravity: gravity, isPictureInPictureSupported: isPictureInPictureSupported)
            }
        case .unknown:
            Color.clear
        }
    }
    ```

2. Run the demo, open an example with the custom player and enable PiP with the dedicated button. PiP is incorrectly exited.

### Library version

0.8.0

### Operating system

iOS 17

### Is there an existing issue for this?

- [X] I have searched existing issues and found no similar bug report.

defagos:
I already opened a branch and updated the demo according to the above report. I have no clue for a solution yet.

defagos:
I think I found a strategy:

- Turn `enabledForInAppPictureInPictureWithCleanup(perform:)` into methods on `VideoView` and `SystemVideoView`. This way we cannot have a dangling modifier without associated view. This should fix the reference counting issue reliably.
- Update our implementation so that it compiles again.
- Maybe our implementation can then be simplified. We namely might be able to get rid of the `PictureInPicture` umbrella object in the process (remains to be investigated).

defagos:
Not sure the above strategy works, as it binds the lifetime of the extended PiP to the lifetime of the video view, thus extended nothing.

Working on another strategy. Just a side note: It seems pauses during PiP restoration might occur when the item is switched during the PiP restoration.

defagos:
The very same issue affects our ability to implement advanced PiP in the `MultiView` example.

PR #698, opened 2023-12-15
Fix Picture in Picture incorrectly stopped when no video view is initially available

defagos:
# Description

The problem reported in the associated issue is related to the fact that no player layer might be initially available. This is something that we cannot control reliably, as implementations might choose to display no video view initially.

I considered two fixes, both as last commits on the associated branch:

- Perform reference counting without checking if a player layer is available.
- Perform reference counting as before when a player layer is available, but perform separate pending reference counting when no player layer is available.

I think the first option should be enough, except if two video views are displayed on a parent view which is enabled for in-app PiP (in which case reference counting might be interleaved).

Once the above fix was made I discovered another issue with the sample code provided in the linked issue. PiP works but is paused when changing the content. It namely appears that updating the player item must be made after PiP will stop method is called. To achieve this result I had to defer setup of the new item, which can only be made by introducing a setup block to the in-app PiP modifier. This block makes it possible to perform the setup after PiP is stopped, with a very small delay required, though.

# Remark

I'll put this PR in review (without auto-merge) but I think we should:

- Discuss the fix first. Implementing in-app PiP for the multiple view demo might help the discussion.
- Test the behavior on an Apple TV device (which I don't have).

# Changes made

- Fix PiP bug. 
- Update the in-app PiP modifier to take a setup closure.
- Remove player layout (makes things more complicated, especially since the system player layout can be tested under the Showcase tab).
- Remove `WillAppearView`, not required anymore.
- Update all demo examples.

# Checklist


Sources/Player/UserInterface/View.swift
```diff
@@ -34,12 +34,9 @@ public extension View {
     /// required by Picture in Picture initiated from the view are not needed anymore.
     ///
     /// - Parameter cleanup: A closure to clean resources associated with the view.
-    func enabledForInAppPictureInPictureWithCleanup(perform cleanup: @escaping () -> Void) -> some View {
+    func enabledForInAppPictureInPictureWithSetup(perform setup: @escaping () -> Void, cleanup: @escaping () -> Void) -> some View {
```
MGaetan89, line 37: Don't you need to update the comment of this method?

waliid:
Thank you for this PR. However, I wonder if we are not going too far.

1. Is it an edge case?
2. Should we try to fix this issue?
3. Shouldn't we always have a `VideoView`, regardless of the media type?

I don't see a situation where we wouldn't want a `VideoView`.
If I want to play content, regardless of its state, I need a `VideoView`.

It seems to me that we are addressing an edge case, but I may be wrong.

defagos:
> 1. Is it an edge case?

If your implementation always has a `VideoView` then, yes, this is an edge case (this is why it was not discovered until now).

> 2. Should we try to fix this issue?
> 3. Shouldn't we always have a `VideoView`, regardless of the media type?

We have no control over how the player will be integrated, especially not if a `VideoView` is guaranteed to be available when the PiP enabled view appears for the first time.

The problem is that clients integrating the library will see PiP fail for no reason. Check the [sample code](https://github.com/SRGSSR/pillarbox-apple/issues/696#issue-2042187883). How is it possible that such a small change can make PiP fail? Clients will never be able to understand and debug the issue.

One possibility was mentioned in the associated issue: Make the modifiers be applied on `VideoView` or `SystemVideoView` itself. This way both can never be separated. But sadly this approach does not work either since it couples the lifetime of PiP too tightly to the video view, leading to other issues.

If we think this is going too far we can:

- Do nothing, but this issue will arise.
- Force a `VideoView` to be always available when PiP reference counting takes place. This could be achieved with an assertion, though in this case there is no guarantee that the issue will be discovered early in client code (not all layout possibilities might be tested well enough in client code).

There are also two implementations possible, as mentioned in the PR description. While the second one is maybe too much, the first one is actually shorter than our initial implementation, since it removes all layer identity tests.

Sources/Player/UserInterface/View.swift
defagos: Fixed, thanks.

defagos:
There is no hurry in merging this PR. I discovered the issue when implementing 360° support (there is a video view, but it cannot be enabled for PiP). I'll have a look at other options.

defagos:
I moved this PR to draft.

I am strongly convinced we have to address this issue. We cannot rely on users correctly inserting a video view in PiP-enabled views, this can break too easily when refactoring SwiftUI code, something quite common (and I was already bitten several times).

The solution of this PR is not good enough, though. It involves API changes (setup phase) which are likely not what we want. I will try to find an elegant solution instead.

defagos:
Found a way to make our current PiP implementation simpler and to solve this issue. This work has been done alongside 360° integration since a complete implementation requires both issues to be solved. I'll open the dedicated PR later today.

This PR can be forgotten. Closed.

PR #711, opened 2024-01-02
Add an access to Pillarbox Web via a WebView

waliid:
# Description

This PR enables us to provide access to the demo app of **Pillarbox Web** through a `WebView`.

# Changes made

- A new showcase has been added, which can be enabled in the device settings.

| Settings | Showcase |
|--------|--------|
| device-settings | showcase | 

# Checklist


Demo/Resources/Settings.bundle/en.lproj/Root.strings
defagos, line 1: You should update this file to match the localizable keys from the `Root.plist` file (you can also remove the comment at the top I guess).

Demo/Sources/Settings/SettingsBundle.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  Copyright (c) SRG SSR. All rights reserved.
+//
+//  License information is available from the LICENSE file.
+//
+
+import Combine
+import Foundation
+
+class SettingsBundle: ObservableObject {
```
defagos, line 10:
```suggestion
final class SettingsBundle: ObservableObject {
```

defagos, line 10:
```suggestion
final class SettingsBundle: ObservableObject {
```

Demo/Sources/Showcase/ShowcaseView.swift
```diff
@@ -348,6 +350,15 @@ struct ShowcaseView: View {
         }
         .sourceCode(of: OptInView.self)
     }
+    @ViewBuilder
```
defagos, line 353:
```suggestion

    @ViewBuilder
```

Demo/Sources/Views/WebView.swift
defagos, line 1: I would suggest we add `~ios` to the Swift file name so that this source file is only compiled for iOS (instead of having a view without purpose on tvOS).

defagos (approved):

defagos:
On iPhone the Done button is currently moved outside the safe area after rotating back from landscape orientation. This is a minor issue IMHO, which does not justify reverting to [UIKit presentation](https://sarunw.com/posts/sfsafariviewcontroller-in-swiftui/).

PR #718, opened 2024-01-05
Implementing Playback Speed Feature on tvOS

waliid:
# Description

This PR allows us to add the playback speed feature on tvOS.

# Changes made

- Added a `Coordinator` to `PictureInPictureSupportingSystemVideoView` for managing the display of playback speed on tvOS.

| Screenshot |
|--------|
| screenshot | 

# Checklist


defagos (commented):
You were right, the code needs to be shared between `BasicSystemVideoView` and `PictureInPictureSupportingVideoView`. 

Suggestion: Maybe you should extract the whole coordinator and make it a tvOS-only type, then use it from both implementations (with no coordiator on iOS).

Sources/Player/UserInterface/AVPlayerViewControllerSpeedCoordinator~tvos.swift
defagos, line 1: The `~tvos` extension is only taken into account in the demo project, not in packages. For consistency with other files in the package I suggest we remove it.

Sources/Player/UserInterface/AVPlayerViewControllerSpeedCoordinator~tvos.swift
```diff
@@ -0,0 +1,75 @@
...
+//
+//  License information is available from the LICENSE file.
+//
+
+import AVKit
+import Combine
+import UIKit
+
+@available(iOS, unavailable)
+class AVPlayerViewControllerSpeedCoordinator {
```
defagos, line 12:
```suggestion
final class AVPlayerViewControllerSpeedCoordinator {
```

Sources/Player/UserInterface/AVPlayerViewControllerSpeedCoordinator~tvos.swift
```diff
@@ -0,0 +1,75 @@
...
+        range: ClosedRange<Float>,
+        speed: Float
+    ) {
+        let speedActions = allowedSpeeds(from: range).sorted().map { allowedSpeed in
+            UIAction(title: String(format: "%g×", allowedSpeed), state: speed == allowedSpeed ? .on : .off) { [weak player] action in
+                player?.playbackSpeed.wrappedValue = allowedSpeed
+                action.state = .on
+            }
+        }
+
```
defagos, line 65:
```suggestion
```

Sources/Player/UserInterface/AVPlayerViewControllerSpeedCoordinator~tvos.swift
```diff
@@ -0,0 +1,75 @@
...
+        Set(
+            AVPlaybackSpeed.systemDefaultSpeeds
+                .map(\.rate)
+                .filter { rate in
+                    range.lowerBound <= rate && rate <= range.upperBound
+                }
+        )
+    }
+
+    private static func configureSpeeds(
```
defagos, line 53:
```suggestion
    static func configureSpeeds(
```

Sources/Player/UserInterface/AVPlayerViewControllerSpeedCoordinator~tvos.swift
```diff
@@ -0,0 +1,75 @@
...
+                Self.configureSpeeds(for: player, controller: controller, range: range, speed: speed.effectiveValue)
+            }
+        }
+        .store(in: &cancellables)
+    }
+}
+
+@available(iOS, unavailable)
+extension AVPlayerViewControllerSpeedCoordinator {
+    private static func allowedSpeeds(from range: ClosedRange<Float>) -> Set<Float> {
```
defagos, line 43:
```suggestion
private extension AVPlayerViewControllerSpeedCoordinator {
    static func allowedSpeeds(from range: ClosedRange<Float>) -> Set<Float> {
```

defagos, line 43:
```suggestion
private extension AVPlayerViewControllerSpeedCoordinator {
    static func allowedSpeeds(from range: ClosedRange<Float>) -> Set<Float> {
```

Sources/Player/UserInterface/AVPlayerViewControllerSpeedCoordinator~tvos.swift
```diff
@@ -0,0 +1,75 @@
...
+        Set(
+            AVPlaybackSpeed.systemDefaultSpeeds
+                .map(\.rate)
+                .filter { rate in
+                    range.lowerBound <= rate && rate <= range.upperBound
+                }
```
defagos, line 49:
```suggestion
                .filter { range.contains($0) }
```

defagos, line 49:
```suggestion
                .filter { range.contains($0) }
```

Sources/Player/UserInterface/PictureInPictureSupportingSystemVideoView.swift
```diff
@@ -23,12 +23,29 @@ struct PictureInPictureSupportingSystemVideoView: UIViewControllerRepresentable
...
+        PictureInPicture.shared.system.relinquish(for: uiViewController)
+    }
+
+    func makeCoordinator() -> AVPlayerViewControllerSpeedCoordinator {
+        .init(
+            player: player,
+            controller: PictureInPicture.shared.system.playerViewController ?? PlayerViewController()
+        )
+    }
+#endif
```
defagos, line 41: Would avoid having two different implementations, rather make the difference with a typealias for the coordinator:

```suggestion
#if os(iOS)
    typealias Coordinator = Void
#else
    typealias Coordinator = AVPlayerViewControllerSpeedCoordinator
#endif

    static func dismantleUIViewController(_ uiViewController: AVPlayerViewController, coordinator: Coordinator) {
        PictureInPicture.shared.system.relinquish(for: uiViewController)
    }

#if os(tvOS)
    func makeCoordinator() -> AVPlayerViewControllerSpeedCoordinator {
        .init(
            player: player,
            controller: PictureInPicture.shared.system.playerViewController ?? PlayerViewController()
        )
    }
#endif
```

defagos (commented):
Warning: With the dual implementation iOS / tvOS the PiP is broken, as we are not restoring a previously available instance.

I'll stop my review here. I think we should avoid duplicating the existing code in the coordinator, this will make things only harder to follow. Maybe we should create the controller in the `make` method and hook it to the coordinator, instead of creating the view controller in the coordinator.

We can talk about this tomorrow, I guess it will be easier.

Sources/Player/UserInterface/AVPlayerViewControllerSpeedCoordinator~tvos.swift
```diff
@@ -0,0 +1,75 @@
...
+    init(player: Player, controller: AVPlayerViewController) {
+        self.player = player
+        self.controller = controller
+        configurePlaybackSpeedPublisher(player: player, controller: controller)
+    }
+
+    private func configurePlaybackSpeedPublisher(player: Player, controller: AVPlayerViewController) {
+        Publishers.CombineLatest(
+            player.playbackSpeedUpdatePublisher(),
+            player.$_playbackSpeed
```
defagos, line 26: `$_playbackSpeed` is derived from `playbackSpeedUpdatePublisher()` so this seems redundant, especially if the goal is to avoid duplication removal. I think we could think of another way to write this publisher.

waliid: What do you think about the following suggestion?

```swift
player.$_playbackSpeed
            .receiveOnMainThread()
            .sink { speed in
                guard let range = speed.range else { return }
                if range == 1...1 {
                    controller.transportBarCustomMenuItems = []
                }
                else {
                    Self.configureSpeeds(for: player, controller: controller, range: range, speed: speed.effectiveValue)
                }
            }
            .store(in: &cancellables)
```

defagos: Would be possible, though I would suggest using the underlying publisher rather than a published property publisher where possible. 

I probably miss the big picture a bit here, though. Maybe it's best we discuss this tomorrow morning so that I can have a deeper understanding of the problems you faced during the implementation.

Sources/Player/UserInterface/AVPlayerViewControllerSpeedCoordinator.swift
```diff
@@ -0,0 +1,70 @@
...
+    private func configurePlaybackSpeedPublisher(player: Player, controller: AVPlayerViewController) {
+        player.playbackSpeedPublisher()
+            .receiveOnMainThread()
+            .sink { speed in
+                guard let range = speed.range else { return }
+                if range == 1...1 {
+                    controller.transportBarCustomMenuItems = []
+                }
+                else {
+                    Self.configureSpeeds(for: player, controller: controller, range: range, speed: speed.effectiveValue)
```
defagos, line 32:
```suggestion
                    controller.transportBarCustomMenuItems = Self.speedMenuItems(for: player, range: range, speed: speed.effectiveValue)
```

Sources/Player/UserInterface/AVPlayerViewControllerSpeedCoordinator.swift
```diff
@@ -0,0 +1,70 @@
...
+        controller.transportBarCustomMenuItems = [
+            UIMenu(
+                title: "Playback Speed",
+                image: UIImage(systemName: "speedometer"),
+                options: [.singleSelection],
+                children: speedActions
+            )
+        ]
+    }
+}
```
defagos, line 70:
```suggestion
    static func speedMenuItems(
        for player: Player,
        range: ClosedRange<Float>,
        speed: Float
    ) -> [UIMenuElement] {
        let speedActions = allowedSpeeds(from: range).sorted().map { allowedSpeed in
            UIAction(title: String(format: "%g×", allowedSpeed), state: speed == allowedSpeed ? .on : .off) { [weak player] action in
                player?.playbackSpeed.wrappedValue = allowedSpeed
                action.state = .on
            }
        }
        return [
            UIMenu(
                title: "Playback Speed",
                image: UIImage(systemName: "speedometer"),
                options: [.singleSelection],
                children: speedActions
            )
        ]
    }
```

Sources/Player/UserInterface/AVPlayerViewControllerSpeedCoordinator.swift
```diff
@@ -0,0 +1,70 @@
...
+        player.playbackSpeedPublisher()
+            .receiveOnMainThread()
+            .sink { speed in
+                guard let range = speed.range else { return }
+                if range == 1...1 {
+                    controller.transportBarCustomMenuItems = []
+                }
+                else {
+                    Self.configureSpeeds(for: player, controller: controller, range: range, speed: speed.effectiveValue)
+                }
```
defagos, line 33:
```suggestion
                if let range = speed.range, range != 1...1 {
                    controller.transportBarCustomMenuItems = Self.speedMenuItems(for: player, range: range, speed: speed.effectiveValue)
                }
                else {
                    controller.transportBarCustomMenuItems = []
                }
```

defagos (approved):

PR #722, opened 2024-01-08
Create a computed property to detect non-zero playback rates

waliid:
# Description

This PR enables us to add a shortcut property to know if the player is playing/active/running.

# Changes made

- Added `isRunning` property.

# Checklist

### DIFF
--- a/Sources/Player/Player.swift
+++ b/Sources/Player/Player.swift
@@ -118,7 +118,9 @@ public final class Player: ObservableObject, Equatable {
     }
 
     /// A Boolean value indicating whether playback is currently playing.
-    public var isRunning: Bool { rate != 0 }
+    public var isRunning: Bool {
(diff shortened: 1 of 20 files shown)


Sources/Player/Player.swift
```diff
@@ -117,6 +117,9 @@ public final class Player: ObservableObject, Equatable {
         }
     }
 
+    /// A Boolean value indicating whether playback is currently playing.
+    public var isRunning: Bool { rate != 0 }
```
defagos, line 121: We usually write methods not as one-liners, if we want to I would suggest we do it everywhere in the codebase, through another PR.

```suggestion
    public var isRunning: Bool {
        rate != 0
    }
```

defagos (approved):

defagos (approved):

PR #746, opened 2024-01-21
Increase playback speed during a long press gesture

waliid:
# Description

This PR enables us updating the speed to 2x when a long press is detected.

# Changes made

- Added `HighSpeedGestureView` detecting when a long press is triggered.
- Added `HighSpeedView` managing the playback, the speeding up.
- Added `HighSpeedCapsule` displaying speed information.
- Added `supportsHighSpeed` view modifier, applying the long press playback speed on the view.

| Not applied | Applied |
|--------|--------|
| no-speed-up | speed-up |

# Checklist


Demo/Sources/Views/LongPressPlaybackView.swift
```diff
@@ -0,0 +1,137 @@
...
+                state = value
+                transaction.animation = .easeIn(duration: 2.0)
+            }
+    }
+}
+
+private struct LongPressPlaybackInfoView: View {
+    var body: some View {
+        HStack {
+            Text("x 2")
```
defagos, line 52:
```suggestion
            Text("×2")
```

Demo/Resources/Localizable.xcstrings
```diff
@@ -193,6 +193,9 @@
     },
     "Web" : {
 
+    },
+    "x 2" : {
```
defagos, line 197:
```suggestion
    "×2" : {
```

defagos (commented):
I suggest we make `PlaybackView.swift` available only to iOS with a `~iOS` prefix and use the preprocessor at the call site (since there is only one). I namely find it strange to have a modifier which implements a whole UI behavior apply on tvOS with no effect.

Demo/Sources/Views/LongPressPlaybackView.swift
```diff
@@ -0,0 +1,137 @@
...
+                .font(.footnote)
+                .bold()
+            HStack(spacing: 3) {
+                ForwardArrowView()
+                ForwardArrowView()
```
defagos, line 57: Likely easier to either:

- [Inline](https://www.hackingwithswift.com/quick-start/swiftui/how-to-insert-images-into-text) image in text directly.
- Use Unicode characters, for example ≫ (U+226B) and ≪(U+226A).

Demo/Sources/Views/LongPressPlaybackView.swift
```diff
@@ -0,0 +1,137 @@
...
+    let action: (_ finished: Bool) -> Void
+
+    var body: some View {
+        content()
+            .simultaneousGesture(longPressGesture())
+            .onChange(of: isLongPressing) { isChanged in
+                timer?.invalidate()
+                if !isChanged {
+                    action(true)
+                } else {
```
defagos, line 26:
```suggestion
                }
                else {
```

Demo/Sources/Views/LongPressPlaybackView.swift
```diff
@@ -0,0 +1,137 @@
...
+        self.minimumPressDuration = minimumPressDuration
+        self.content = content
+        self.action = action
+    }
+
+    private func longPressGesture() -> some Gesture {
+        LongPressGesture(minimumDuration: .infinity, maximumDistance: .infinity)
+            .updating($isLongPressing) { value, state, transaction in
+                state = value
+                transaction.animation = .easeIn(duration: 2.0)
```
defagos, line 44: Not seing where a 2-second animation takes place. Are animations properly forwarded? If an animation is desired, isn't 2 seconds a bit long?

Demo/Sources/Views/LongPressPlaybackView.swift
waliid: I have chosen the inline approach.

Demo/Sources/Views/LongPressPlaybackView.swift
```diff
@@ -0,0 +1,137 @@
...
+    let minimumPressDuration: TimeInterval
+    @ViewBuilder let content: () -> Content
+    let action: (_ finished: Bool) -> Void
+
+    var body: some View {
+        content()
+            .simultaneousGesture(longPressGesture())
+            .onChange(of: isLongPressing) { isChanged in
+                timer?.invalidate()
+                if !isChanged {
```
defagos, line 24:
```suggestion
            .onChange(of: isLongPressing) { isPressing in
                timer?.invalidate()
                if !isPressing {
```

Demo/Sources/Views/LongPressPlaybackView.swift
```diff
@@ -0,0 +1,137 @@
...
+import PillarboxPlayer
+import SwiftUI
+
+#if os(iOS)
+private struct LongPressView<Content>: View where Content: View {
+    @GestureState private var isLongPressing = false
+    @State private var timer: Timer?
+
+    let minimumPressDuration: TimeInterval
+    @ViewBuilder let content: () -> Content
```
defagos, line 16:
```suggestion
    let content: () -> Content
```

`@ViewBuilder` is superfluous when the initializer has been explicitly implemented if I am not mistaken.

Demo/Sources/Views/LongPressPlaybackView.swift
```diff
@@ -0,0 +1,137 @@
...
+    private var canSpeedUp: Bool {
+        player.playbackSpeedRange.contains(2)
+    }
+    private var isPlaying: Bool {
+        playbackState == .playing
+    }
+
+    let minimumPressDuration: TimeInterval
+    let player: Player
+    @ViewBuilder let content: () -> Content
```
defagos, line 91:
```suggestion
    let content: () -> Content
```

Same as above.

Demo/Sources/Views/LongPressPlaybackView.swift
waliid: The animation is slightly perceptible during use. The switch between two speeds (current speed to 2x) seems to be smoother.

defagos (commented):
It would likely also be great if controls where hidden when entering high-speed mode, as in YouTube. It is likely that the intention is to see the content at higher speed, so always hiding the controls makes sense IMHO.

Demo/Sources/Views/LongPressPlaybackView.swift
```diff
@@ -0,0 +1,137 @@
...
+            .resizable()
+            .aspectRatio(contentMode: .fit)
+            .frame(width: 10)
+    }
+}
+
+struct LongPressPlaybackView<Content>: View where Content: View {
+    @State private var speed: Float?
+    @State private var isFinished = true
+    @State private var isBusy = false
```
defagos, line 79: Is busy state observation really essential for correct behavior?

Demo/Sources/Views/LongPressPlaybackView.swift
defagos: With a duration of 2 seconds it should be clearly noticeable, though. I removed it and observed no noticeable change. Is the animation really there?

Demo/Sources/Views/LongPressPlaybackView.swift
```diff
@@ -0,0 +1,137 @@
...
+            }
+            isFinished = finished
+        }
+        .overlay {
+            if !isFinished, !isBusy, isPlaying, canSpeedUp {
+                LongPressPlaybackInfoView()
+                    .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .top)
+                    .padding(.top)
+            }
+        }
```
defagos, line 109: Prefer using overlay alignment directly. The padding direction is also superfluous in this case (views take the space they need in SwiftUI anyway).

```suggestion
        .overlay(alignment: .top) {
            if !isFinished, isPlaying, canSpeedUp {
                LongPressPlaybackInfoView()
                    .padding()
            }
        }
```

Demo/Sources/Views/LongPressPlaybackView.swift
waliid: If we want to avoid displaying the bubble when the player cannot speed up, then yes. I find it strange to show the bubble to the user when speeding up does nothing.

waliid:
> It would likely also be great if controls where hidden when entering high-speed mode, as in YouTube (only for video, of course). It is likely that the intention is to see the content at higher speed, so always hiding the controls makes sense IMHO.

We could probably do it in another story.

defagos (approved):

PR #769, opened 2024-02-13
Improve Control Center display in presence of errors

waliid:
# Description

Self-explanatory.

# Changes made

- Additional error examples have been included.
- The method `replaceItems` has been updated to clear all items when the first queue item encounters an error.
- The `Current` object has been deleted.
- `ItemState` has been renamed to `ItemStatus`.
- New objects, namely `Queue`, `QueueElement`, `QueueUpdate`, `QueueBuffer`, and `ItemState`, have been introduced.
- `ItemUpdatePublisher` has been substituted with `queuePublisher`.
- The order configuration of publishers has been rearranged.
- The use of `currentPublisher` has been replaced with the utilization of `queuePublisher`.
- A new method, `nowPlayingInfo(with error: NSError)`, has been introduced.
- The `skipBackwardCommand`, `skipForwardCommand`, and `changePlaybackPositionCommand` are deactivated if an error occurs.
- The Control Center description is populated with an error message if necessary.

# Checklist


Sources/Player/Player+ControlCenter.swift
```diff
@@ -31,13 +31,22 @@ extension Player {
...
         currentPublisher()
-            .map { current in
+            .map { [queuePlayer] current in
                 guard let current else {
                     return Just(NowPlayingInfo()).eraseToAnyPublisher()
                 }
-                return current.item.$asset
-                    .filter { !$0.resource.isLoading }
-                    .compactMap { $0.nowPlayingInfo() }
+                return Publishers.CombineLatest(current.item.$asset, queuePlayer.errorPublisher())
```
defagos, line 38: Maybe we could use `itemUpdatePublisher` directly? We would be able to:

- Access the stored item list.
- Get the current `AVPlayerItem` and subscribe to its errors (without having to capture `queuePlayer`).

Tests/PlayerTests/Publishers/NowPlayingInfoMetadataPublisherTests.swift
```diff
@@ -110,4 +110,14 @@ final class NowPlayingInfoMetadataPublisherTests: TestCase {
             player.play()
         }
     }
+
+    func testErrorMetadata() {
+        let player = Player(item: .simple(url: Stream.unavailable.url))
+        expectAtLeastSimilarPublished(
+            values: [[MPMediaItemPropertyTitle: "\((PlayerError.resourceNotFound.underlyingErrors.first! as NSError).userInfo["NSDescription"] as! String)"]],
```
defagos, line 117: Maybe we could still keep the content title and display the error as subtitle?

defagos (commented):
If we display failed items in the Control Center we should maybe disable play / pause / toggle play / pause commands (and possibly other), just keeping skip forward and backward commands active.

Sources/Player/Player+ControlCenter.swift
waliid: Unfortunately, we can't use `itemUpdatePublisher` due to the transition from a loading asset to a loaded asset. By using `itemUpdatePublisher`, we are only listening to the transition between items and not between the different states of an item.

defagos (approved):

PR #781, opened 2024-02-29
Avoid unnecessary Commanders Act seek / play transitions

defagos:
# Description

This PR fixes superfluous Commanders Act seek / play transitions seen when briefly interacting with a progress tracker. Those could be easily be reproduced as follows:

1. Open a URN example from the demo, monitoring Commanders Act events with a proxy.
2. Touch down on the progress bar **without** moving the finger sideways.
3. Release the finger. A seek / play transition could be observed.

# Changes made

- Self-explanatory.

# Checklist

### DIFF
--- a/Sources/Player/UserInterface/ProgressTracker.swift
+++ b/Sources/Player/UserInterface/ProgressTracker.swift
@@ -58,8 +58,8 @@ public final class ProgressTracker: ObservableObject {
                 pausePlaybackIfNeeded(with: player)
             }
             else {
-                resumePlaybackIfNeeded(with: player)
                 seek(to: progress, optimal: false)
+                resumePlaybackIfNeeded(with: player)
             }
         }
     }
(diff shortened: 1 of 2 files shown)


defagos:
After discussion with @waliid we removed the test corresponding to the code change (simple line inversion). Two reasons:

- The test is made in PillarboxAnalytics but should be made in PillarboxPlayer instead (since we really changed a player behavior, and knowing we will separate PillarboxPlayer in a near future).
- If written in PillarboxPlayer test suites the test is likely a bit too much coupled to the implementation.

For the sake of history here is the test that was made to ensure that no unnecessary seek / play transition was made during brief interactions:

```swift
func testCleanEventsDuringBriefProgressTrackerInteraction() {
    let player = Player(item: .simple(
        url: Stream.onDemand.url,
        trackerAdapters: [
            CommandersActTracker.adapter()
        ]
    ))
    player.play()
    expect(player.playbackState).toEventually(equal(.playing))

    let progressTracker = ProgressTracker(interval: CMTime(value: 1, timescale: 4))
    progressTracker.player = player
    expectHits(.pause(), .play(), during: .seconds(1)) {
        progressTracker.isInteracting = true
        progressTracker.isInteracting = false
    }
}
```

waliid (approved):

Issue #785, opened 2024-03-05
Control Center metadata may disappear on current item deletion


defagos:
### Description of the problem

When deleting the item currently being played from a playlist, the metadata displayed in the Control Center might vanish (time information stays correct, though).

### Reproducibility

Always

### Steps to reproduce

1. Open a URN playlist example.
2. Tap on some item to play it (in the middle of the list).
3. Tap on an item located **before the item you just selected** (this is why we started in the middle).
4. Delete the current item.
5. Check the Control Center. The metadata is missing.

This is really an issue in our code since the metadata attached to the now playing center is missing when logged to the console.

### Library version

0.10.0

### Operating system

iOS 17

### Is there an existing issue for this?

- [X] I have searched existing issues and found no similar bug report.

defagos:
This is a regression introduced in #769.

defagos:
The issue is related to ordering of registrations in multicasting:

- I reproduced the queue update mechanism we have in a small project and introduced replay again. By simply adding dummy registrations at different places relative to the queue update publisher registration (all bound to the same shared publisher) I could change the obtained output.
- Simply reversing the order of iteration in our `ReplaySubject` implementation makes the behavior different (bad since the player then freezes).

defagos:
There is a clear difference in behavior between the following cases:

- Direct update of the player items, e.g. via advance / return APIs or direct index selection.
- Indirect update of the player items, e.g. via reordering of the items.

The first is mostly related to current item updates (though lazy loading can later lead to asset updates), while the second is related to stored item updates (and thus to asset updates).

It also means we can simply reproduce the very same issue as follows:

1. Open a playlist demo.
2. Reorder 2 random items other than the one being played.
3. Delete the item being played.

Maybe careful analysis of what happens in this simpler case might provide useful insights into the actual problem.

defagos:
I could properly debug and understand the origin of the issue. As initially suspected this is because of an update loop involving `ReplaySubject`.

Let us assume we have the following a publisher which shares its values to multiple suscribers with a `multicast`. A `ReplaySubject` is inserted in the middle so that all subscribers can receive recent values upon subscription:

```
   ┌──────────────────────────┐    ┌────────────────────────┐    ┌─────────────────────┐   
┌─▶│        Publisher         │───▶│     ReplaySubject      │─┬─▶│    Subscriber 1     │   
│  └──────────────────────────┘    └────────────────────────┘ │  └─────────────────────┘   
[...]

PR #786, opened 2024-03-10
Fix Control Center metadata sometimes disappearing after item deletion

defagos:
# Description

This PR fixes an issue with Control Center metadata sometimes disappearing after item deletion, e.g. after reshuffling the item after the current one with another down in a playlist, then deleting the current item.

The issue was located in our `ReplaySubject` implementation (similar implementations like CombineExt or Entwine have the same issue). See #785 for an explanation of the issue and how to solve it.

# Changes made

- Fix `ReplaySubject` implementation to support in-order value processing in multicast configurations where some subscriber might trigger a new publisher emission on the same thread.
- Make `ReplaySubject` public (just an oversight).
- Extracted `QueuePlayer`-related types into separate files (general improvement).

# Checklist

### DIFF
--- a/Sources/Core/ReplaySubscription.swift
+++ b/Sources/Core/ReplaySubscription.swift
@@ -10,6 +10,7 @@ import Foundation
 final class ReplaySubscription<Output, Failure>: Subscription where Failure: Error {
     private var subscriber: AnySubscriber<Output, Failure>?
     private var buffer = DemandBuffer<Output>()
+    private var pendingValues: [Output] = []
 
     init<S>(subscriber: S, values: [Output]) where S: Subscriber, S.Input == Output, S.Failure == Failure {
         self.subscriber = AnySubscriber(subscriber)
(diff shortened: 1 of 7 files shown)


Sources/Core/ReplaySubscription.swift
```diff
@@ -23,8 +24,14 @@ final class ReplaySubscription<Output, Failure>: Subscription where Failure: Err
...
-    func send(_ value: Output) {
-        process(buffer.append(value))
+    func append(_ value: Output) {
+        pendingValues += buffer.append(value)
+    }
+
+    func send() {
+        let values = pendingValues
+        pendingValues = []
+        process(values)
```
waliid, line 34:
```suggestion
        process(pendingValues)
        pendingValues = []
```

defagos: Would of course be logical but does not work, see 2 in this [comment](https://github.com/SRGSSR/pillarbox-apple/issues/785#issuecomment-1987115101).

waliid (approved):

PR #866, opened 2024-05-03
Display artwork image for audio

defagos:
# Description

This PR updates the custom player UI to display an artwork image, if available, when audio is played.

| Before | After |
|--------|--------|
| Before | After | 

# Changes made

- Display image when available.
- Remove audio image placeholder when no image is available.
- Slightly increase control overlay opacity.

# Checklist

### DIFF
--- a/Demo/Sources/Players/PlaybackView.swift
+++ b/Demo/Sources/Players/PlaybackView.swift
@@ -201,10 +201,23 @@ private struct MainView: View {
     }
 
     @ViewBuilder
+    private func audioArtwork() -> some View {
+        if let image = player.metadata.image {
+            Image(uiImage: image)
+                .resizable()
+                .aspectRatio(contentMode: .fit)
+                .frame(maxWidth: .infinity, maxHeight: .infinity)
+        }
+        else {
+            image(name: "music.note.tv.fill")
+        }
+    }
+
+    @ViewBuilder
     private func video() -> some View {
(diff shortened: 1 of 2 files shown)


Demo/Sources/Players/PlaybackView.swift
```diff
@@ -200,11 +204,19 @@ private struct MainView: View {
...
+            .frame(maxWidth: .infinity, maxHeight: .infinity)
+    }
+
     @ViewBuilder
     private func video() -> some View {
         ZStack {
-            if player.mediaType == .audio {
-                image(name: "music.note.tv.fill")
+            if player.mediaType == .audio, let image {
+               artwork(for: image)
```
waliid, line 219:
```suggestion
				if player.mediaType == .audio {
					if let image {
						artwork(for: image)
					}
					else {
						image(name: "music.note.tv.fill")					}
```

waliid (approved):

Demo/Sources/Players/PlaybackView.swift
defagos: I had initially written the code this way, but I thought the result was better without the music note placeholder after testing it.

We can of course keep it if you think the result is better.

waliid: Without placeholder is also fine.

waliid (approved):

PR #881, opened 2024-05-15
Fix undesired animation

waliid:
# Description

This PR allows us to fix an undesired animation on the metadata view.

# Changes made

- Renames some computed variables.
- Uses same logic for opacities.
- Hides bottom buttons when the UI should be hidden.

# Checklist

### DIFF
--- a/Demo/Sources/Players/PlaybackView.swift
+++ b/Demo/Sources/Players/PlaybackView.swift
@@ -25,16 +25,16 @@ private struct MainView: View {
     @State private var selectedGravity: AVLayerVideoGravity = .resizeAspect
     @State private var isInteracting = false
 
-    private var areControlsAlwaysVisible: Bool {
-        player.isExternalPlaybackActive || player.mediaType != .video
-    }
-
     private var prioritizesVideoDevices: Bool {
         player.mediaType == .video
     }
 
     private var shouldHideInterface: Bool {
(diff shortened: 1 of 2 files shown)


Demo/Sources/Players/PlaybackView.swift
```diff
@@ -185,7 +191,7 @@ private struct MainView: View {
                 VolumeButton(player: player)
             }
         }
-        .opacity(isUserInterfaceHidden ? 0 : 1)
+        .opacity(shouldHideInterface ? 0 : 1)
```
defagos, line 194: The change is not animated. Maybe:

```suggestion
        .opacity(shouldHideInterface ? 0 : 1)
        .animation(.defaultLinear, value: shouldHideInterface)
```

defagos (approved):
Two other remarks while I was looking at the code (not related to your changes):

- `.linear(duration: 0.2)` should just be `.defaultLinear`.
- The `Self` in `static let defaultLinear = Self.linear(duration: 0.2)` can be removed.

Would you mind doing these changes and commit them? Thanks in advance!

Demo/Sources/Players/PlaybackView.swift
waliid: Should be fixed.

defagos: Wonderful, thanks!

defagos (approved):

PR #891, opened 2024-05-18
Improve server support in Core Business

defagos:
# Description

This PR improves server support in Core Business. It also enhances server selection in the demo to support SAM as well as simulated locations in CH or abroad.

# Changes made

- Make it possible to use any server conforming to the media composition data format and image resizing. 
- Remove `.mmf` from Core Business server list (development tool). This is a breaking API change.
- Add support for SAM (prod, stage, test).
- Add support for simulated CH and WW locations for IL services (helpful to test geoblocking behavior).

# Checklist

### DIFF
--- a/Demo/Sources/Model/ServerSetting.swift
+++ b/Demo/Sources/Model/ServerSetting.swift
@@ -17,9 +17,9 @@ enum ServerSetting: Int, CaseIterable {
     case ilStageCH
     case ilTestCH
 
-    case productionWW
-    case stageWW
-    case testWW
+    case ilProductionWW
+    case ilStageWW
+    case ilTestWW
 
     case samProduction
     case samStage
@@ -41,11 +41,11 @@ enum ServerSetting: Int, CaseIterable {
             return "IL Stage - CH"
         case .ilTestCH:
             return "IL Test - CH"
-        case .productionWW:
+        case .ilProductionWW:
             return "IL Production - WW"
-        case .stageWW:
+        case .ilStageWW:
             return "IL Stage - WW"
-        case .testWW:
+        case .ilTestWW:
             return "IL Test - WW"
         case .samProduction:
             return "SAM Production"
@@ -66,11 +66,11 @@ enum ServerSetting: Int, CaseIterable {
 
     private var baseUrl: URL {
         switch self {
-        case .ilProduction, .ilProductionCH, .productionWW:
+        case .ilProduction, .ilProductionCH, .ilProductionWW:
             return SRGIntegrationLayerProductionServiceURL()
-        case .ilStage, .ilStageCH, .stageWW:
+        case .ilStage, .ilStageCH, .ilStageWW:
             return SRGIntegrationLayerStagingServiceURL()
-        case .ilTest, .ilTestCH, .testWW:
+        case .ilTest, .ilTestCH, .ilTestWW:
             return SRGIntegrationLayerTestServiceURL()
         case .samProduction:
             return SRGIntegrationLayerProductionServiceURL().appending(component: "sam")
@@ -94,7 +94,7 @@ enum ServerSetting: Int, CaseIterable {
         switch self {
         case .ilProductionCH, .ilStageCH, .ilTestCH:
             return [URLQueryItem(name: "forceLocation", value: "CH")]
-        case .productionWW, .stageWW, .testWW:
+        case .ilProductionWW, .ilStageWW, .ilTestWW:
             return [URLQueryItem(name: "forceLocation", value: "WW")]
         case .samProduction, .samStage, .samTest:
             return [URLQueryItem(name: "forceSAM", value: "true")]
--- a/Demo/Sources/Model/ServerSetting.swift
+++ b/Demo/Sources/Model/ServerSetting.swift
@@ -36,17 +36,17 @@ enum ServerSetting: Int, CaseIterable {
         case .test:
             return "Test"
         case .productionCH:
-            return "Production (CH)"
+            return "Production - CH"
         case .stageCH:
-            return "Stage (CH)"
+            return "Stage - CH"
         case .testCH:
-            return "Test (CH)"
+            return "Test - CH"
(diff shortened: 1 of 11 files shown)


Demo/Sources/Model/ServerSetting.swift
```diff
@@ -12,6 +12,19 @@ enum ServerSetting: Int, CaseIterable {
...
+    case stageCH
+    case testCH
+
+    case productionWW
+    case stageWW
+    case testWW
+
+    case samProduction
+    case samStage
+    case samTest
```
waliid, line 26: As we prefix these cases with `sam`, should we also prefix the original cases with `il`?

Sources/CoreBusiness/DataProvider/Server.swift
```diff
@@ -7,29 +7,102 @@
...
-        case .stage:
-            return URL(string: "https://il-stage.srgssr.ch")!
-        case .test:
-            return URL(string: "https://il-test.srgssr.ch")!
-        case .mmf:
-            return URL(string: "https://play-mmf.herokuapp.com")!
+    public static let test = custom(baseUrl: URL(string: "https://il-test.srgssr.ch")!)
+
+    private let requestBuilder: (String) -> URLRequest
+    private let resizedImageUrlBuilder: (URL, ImageWidth) -> URL
```
waliid, line 27: These two closures seem to be used many times, so my suggestion is to use type aliases.

```swift
public typealias RequestBuilder = (String) -> URLRequest
public typealias UrlBuilder = (URL, ImageWidth) -> URL
```

Sources/CoreBusiness/DataProvider/Server.swift
```diff
@@ -7,29 +7,102 @@
...
+
+    /// Custom environment.
+    ///
+    /// - Parameters:
+    ///   - baseUrl: The base URL of the server.
+    ///   - queryItems: Additional query items to use.
+    ///
+    /// Useful for servers which can exactly pose as SRG SSR servers and deliver the same playback metadata format and
+    /// image scaling capabilities.
+    public static func custom(baseUrl: URL, queryItems: [URLQueryItem] = []) -> Self {
```
waliid, line 61: What do you think about renaming `custom` by `urn`?

```suggestion
    public static func urn(baseUrl: URL, queryItems: [URLQueryItem] = []) -> Self {
```

Sources/CoreBusiness/DataProvider/Server.swift
```diff
@@ -7,29 +7,102 @@
...
+        .custom { urn in
+            standardRequest(forUrn: urn, baseUrl: baseUrl, queryItems: queryItems)
+        } resizedImageUrlBuilder: { url, width in
+            standardResizedImageUrl(url, width: width, baseUrl: baseUrl, queryItems: queryItems)
+        }
+    }
+
+    private static func standardRequest(forUrn urn: String, baseUrl: URL, queryItems: [URLQueryItem]) -> URLRequest {
```
waliid, line 69:
```suggestion
    private static func request(forUrn urn: String, baseUrl: URL, queryItems: [URLQueryItem]) -> URLRequest {
```

Sources/CoreBusiness/DataProvider/Server.swift
```diff
@@ -7,29 +7,102 @@
...
+            return .init(url: url)
         }
+        components.queryItems = queryItems + [
+            URLQueryItem(name: "onlyChapters", value: "true"),
+            URLQueryItem(name: "vector", value: vector)
+        ]
+        return .init(url: components.url ?? url)
+    }
+
+    private static func standardResizedImageUrl(_ url: URL, width: ImageWidth, baseUrl: URL, queryItems: [URLQueryItem]) -> URL {
```
waliid, line 81:
```suggestion
    private static func resizedImageUrl(_ url: URL, width: ImageWidth, baseUrl: URL, queryItems: [URLQueryItem]) -> URL {
```

Sources/CoreBusiness/DataProvider/Server.swift
defagos: Applied, thanks.

Sources/CoreBusiness/DataProvider/Server.swift
defagos: Applied, thanks.

Sources/CoreBusiness/DataProvider/Server.swift
defagos: Would be misleading IMHO. URN is just an identifier for content on a server, not the server itself. Maybe `custom` is not the proper word for it but we need a way to define custom servers which behave like an IntegrationLayer (e.g. MMF or SAM).

waliid: Should we rename the urn closure parameter by identifier then?

Demo/Sources/Model/ServerSetting.swift
defagos: Done.

Sources/CoreBusiness/DataProvider/Server.swift
defagos: I removed APIs for free construction with closures. Was overkill.

Sources/CoreBusiness/DataProvider/Server.swift
defagos: I exposed the public initializer instead.

defagos:
I revisited my original implementation which was providing closures to make it possible to build media composition requests and custom image resizing to make it simpler. In practice having a custom service posing as the IL is only useful internally anyway and requires deep understanding of endpoints and metadata formats.

We currently have SAM and MMF posing as the IL, as well as specific ways to simulate a user abroad. Those are implemented in the demo (and thus not provided in the Core Business public API). When SAM is released this mechanism will not be useful anymore anyway, so I guess it is better to keep things simple.

waliid (approved):

PR #898, opened 2024-05-28
Add basic keyboard shortcut support

defagos:
# Description

This PR introduces a few obvious keyboard shortcuts which make the user experience better on iPad or macOS (as iPad app):

- Space to toggle play / pause.
- `S` and `D` to for skips.
- `F` to toggle full screen.

# Changes made

Self-explanatory.

# Checklist

### DIFF
--- a/Demo/Sources/Model/Template.swift
+++ b/Demo/Sources/Model/Template.swift
@@ -143,7 +143,7 @@ enum URLTemplate {
         title: "Brain Farm Skate Phantom Flex",
         subtitle: "4K video",
         imageUrl: "https://i.ytimg.com/vi/d4_96ZWu3Vk/maxresdefault.jpg",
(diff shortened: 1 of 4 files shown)


waliid:
What do you think about adding a shortcut `t` to quickly `type` into the text field?

defagos:
Sure, but maybe something looking like the address bar shortcut of a web browser would be more intuitive?

defagos:
Arrow keys would have been more natural and consistent with the standard player UX but they cannot be properly setup with SwiftUI API yet, as `keyboardShortcut` modifiers do not let system priority be overridden, unlike `UIKeyCommand` allows with its `wantsPriorityOverSystemBehavior` property (which we are using in Letterbox).

I reported an API improvement suggestion as FB13816367.

waliid (approved):

Issue #904, opened 2024-06-05
Metrics collector


defagos:
**As a stream delivery team member** I want to be able to gather metrics for representation through graphs.

# Acceptance criteria

- A `MetricsCollector` is provided that gathers real-time or significant metrics, publishing updates automatically, and that can be bound to any player.
- The framework provides a ready to use collector graphical view that makes it easy to understand metrics in any app.
- The demo provides a way to enable graphical view of any playback session.

# Tasks

- [x] Implement `MetricsCollector`.
- [x] Implement a dedicated standard debugging view that displays gathered metrics and numeric values for current metrics.
- [x] Add way to display metrics in the custom player.

PR #912, opened 2024-06-14
Gather player metrics

defagos:
# Description

This PR gathers player metrics by reading player item access logs.

# Changes made

- Add `Metrics` public type, providing information (server, uri, etc.) as well as total values (e.g. number of stalls) and increments between consecutive measurements.
- Store consolidated uninterrupted session metrics in properties.
- Provide accessor for current metrics retrieval at any time.
- Add publisher periodically publishing metrics.
- Update non-working SWI example.

# Checklist


Sources/Player/Player+Metrics.swift
```diff
@@ -0,0 +1,40 @@
...
+                guard let item else { return Just(nil).eraseToAnyPublisher() }
+                return Publishers.PeriodicTimePublisher(for: queuePlayer, interval: interval, queue: kMetricsQueue)
+                    .scan(MetricsState.empty) { initial, next in
+                        initial.updated(with: item.accessLog(), at: next) ?? initial
+                    }
+                    .removeDuplicates()
+                    .withPrevious(MetricsState.empty)
+                    .map { $0.current.metrics(from: $0.previous) }
+                    .eraseToAnyPublisher()
+            }
```
waliid, line 34: I wanted to simplify do that: 

```swift
            .compactMap { $0 }
            .map { [queuePlayer] item -> AnyPublisher<Metrics?, Never> in
                Publishers.PeriodicTimePublisher(for: queuePlayer, interval: interval, queue: kMetricsQueue)
                    .scan(MetricsState.empty) { initial, next in
                        initial.updated(with: item.accessLog(), at: next) ?? initial
                    }
                    .removeDuplicates()
                    .withPrevious(MetricsState.empty)
                    .map { $0.current.metrics(from: $0.previous) }
                    .eraseToAnyPublisher()
            }
```

The code above leads to unexpected metrics (increments with negative values).
I tried to write a test, but unfortunately I failed.

Sources/Player/Metrics/AccessLog.swift
```diff
@@ -0,0 +1,33 @@
...
+    init(events: [AccessLogEvent?], after date: Date?) {
+        closedEvents = Array(events.prefix(max(events.count - 1, 0))).compactMap { event in
+            Self.event(event, after: date)
+        }
+        if let lastEvent = events.last {
+            openEvent = Self.event(lastEvent, after: date)
+        }
+        else {
+            openEvent = nil
+        }
```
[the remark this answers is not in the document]
defagos: Not sure this helps in keeping things readable here. The double optional is quite strange here, usually we would expect `Self.event(events.last, after: date)` which of course here is not valid.

Sources/Player/Player+Metrics.swift
defagos: The `compactMap` is probably not appropriate here. If we have no item we want to have no metrics, so it is quite natural to return `nil` here instead of inhbiting the update because of the `compactMap`.

Sources/Player/Metrics/AccessLog.swift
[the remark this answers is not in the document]
waliid: I see. Other suggestion: 

```suggestion
        closedEvents = Array(events.prefix(max(events.count - 1, 0))).compactMap { Self.event($0, after: date) }
        let lastEvent = events.last?.flatMap(\.self)
        openEvent = Self.event(lastEvent, after: date)
```

Sources/Player/Player+Metrics.swift
waliid: What do you think about adding the following test to avoid any errors in the future?

```swift
func testMetricsWithAnEmptyPlayer() {
    let player = Player()
    expectAtLeastEqualPublished(
        values: [nil],
        from: player.periodicMetricsPublisher(forInterval: .init(value: 1, timescale: 1)),
        timeout: .seconds(1)
    )
}
```

defagos: Sure. I also introduced similar tests on a development branch for #904.

defagos: Done. I also added a basic test with a playing player.

Sources/Player/Metrics/AccessLog.swift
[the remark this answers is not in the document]
waliid: Did you see my last suggestion?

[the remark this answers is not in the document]
defagos: Nope, but as with [the other suggestion](https://github.com/SRGSSR/pillarbox-apple/pull/912#discussion_r1640359995) I think that the shortened code does not translate into more readability.

waliid (approved):

defagos:
@waliid Thanks for approving the PR as it is. We can still discuss any coding style improvements tomorrow and make changes on future branches.

PR #922, opened 2024-07-05
Make metric event API more ergonomic

defagos:
# Description

This PR improves the metric event API to associate events with the current item in a more meaningful and ergonomic way.

# Changes made

- Update `PlayerItemTracker` protocol to receive events one at a time. A tracker is associated with a single item anyway so being able to receive events one at a time is more natural.
- Add shared `metricEventsPublisher` that publishes events for the item currently being played. This way the event history can be fetched at any time without requiring subscription from the very beginning of the playback.
- Use `queuePublisher` as shared common current item source to improve overall consistency. 
- Update `MetricsCollector` to provide events as well.
- Improve metrics view, displaying startup times for the current item.

# Checklist


waliid (approved):

Tests/PlayerTests/Tools/Similarity.swift
```diff
@@ -45,6 +45,19 @@ extension NowPlaying.Info: Similar {
...
+extension MetricEvent: Similar {
+    public static func ~~ (lhs: MetricEvent, rhs: MetricEvent) -> Bool {
+        switch (lhs.kind, rhs.kind) {
+        case (.assetLoading, .assetLoading):
+            return true
+        case (.resourceLoading, .resourceLoading):
+            return true
+        default:
+            return false
+        }
```
waliid, line 57:
```suggestion
        switch (lhs.kind, rhs.kind) {
        case (.assetLoading, .assetLoading), (.resourceLoading, .resourceLoading):
            return true
        default:
            return false
        }
```

waliid (approved):

waliid (approved):

PR #938, opened 2024-07-15
Capture stalls in QoS metrics

defagos:
# Description

This PR logs stalls to our metric event log.

# Changes made

- Add new measurement method to calculate date intervals based on start and end conditions.
- Provide `isStalled` state in item properties.
- Add new metric events for stalls (start / end).
- Measure date intervals when `isStalled` changes from `true` to `false`.

# Checklist

### DIFF
--- a/Sources/Core/Publisher.swift
+++ b/Sources/Core/Publisher.swift
@@ -255,4 +255,35 @@ public extension Publisher {
         )
         .eraseToAnyPublisher()
     }
+
+    /// Measure date intervals, starting after some condition is satisfied and ending when another one is satisfied.
+    ///
+    /// - Parameters
+    ///    - measure: A closure called for each output delivered upstream.
+    ///    - startCondition: A closure called to determine when a measurement must start.
+    ///    - endCondition: A closure called to to determine when a measurement must stop.
+    ///
+    /// A measurement is made for each pair of matching conditions.
+    func measureDateInterval(
+        perform measure: @escaping (DateInterval) -> Void,
+        after startCondition: @escaping (Output) -> Bool,
+        when endCondition: @escaping (Output) -> Bool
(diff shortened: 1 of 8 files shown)


Sources/Core/Publisher.swift
```diff
@@ -249,4 +255,35 @@ public extension Publisher {
...
+    /// - Parameters
+    ///    - measure: A closure called for each output delivered upstream.
+    ///    - startCondition: A closure called to determine when a measurement must start.
+    ///    - endCondition: A closure called to to determine when a measurement must stop.
+    ///
+    /// A measurement is made for each pair of matching conditions.
+    func measureDateInterval(
+        perform measure: @escaping (DateInterval) -> Void,
+        after startCondition: @escaping (Output) -> Bool,
+        when endCondition: @escaping (Output) -> Bool
```
waliid, line 270:
```suggestion
        until endCondition: @escaping (Output) -> Bool
```

Tests/CoreTests/MeasurePublisherTests.swift
```diff
@@ -142,4 +142,34 @@ final class MeasurePublisherTests: XCTestCase {
             expect(duration).to(beCloseTo(expected, within: 0.1))
         }
     }
+
+    func testAfterWhen() {
```
waliid:
```suggestion
    func testAfterUntil() {
```

waliid (approved):

Sources/Core/Publisher.swift
defagos: Thanks, I propose `measureEachDateInterval(perform:after:until:)` to make it explicit that measurements always take place between pairs of conditions (and not only the first time).

waliid (approved):

PR #950, opened 2024-07-31
Revisit player item tracking

waliid:
# Description

This PR allows us to improve the player item tracker implementation. Indeed, with the previous implementation, we weren't able to get the player time when the tracker was disabled. 
This led introducing a complex logic for some tracker to determine the player time.
This PR solve this issue by directly passing the player properties when the tracker is disabled.

# Changes made

- Some schemes have been updated by a **Pillarbox** prefix.
- The `PlayerItemTracker` contract has been updated. 
	- The `Player` is no longer passed as a parameter of `enable(for:)` method to avoid shortcuts that could lead to unexpected behavior if the player is not used correctly. Instead, we provide an `AVPlayer` which is the low-level player that can be used for various purposes (e.g. provided to a third-party API like **Youbora**).
	- The `disable` method now includes a new `PlayerProperties` parameter. This can be useful for retrieving the player time, for example.
- The `OrderedCollections` dependency has been removed from `PillarboxCore`.
- `CommandersAct` tracker has been refactored to avoid interpolating the player time when the tracker is disabled.
- A `Stopwatch` object has been introduced to determine how much time a content has been played. It's only used for `CommandersAct` tracking purposes for the moment.
- The method `measureDateInterval`, which measures the interval between consecutive events, now directly uses the API provided by Apple.
- The `CurrentTracker` has been renamed to `Tracker`.
- The `Tracker` provides a metrics `metricEventsPublisher` to get all metric events related to the item being played.
- A potential flaky test (`testFailingUrl()`) has been fixed.

# Checklist


defagos:
The `Tracker` implementation was still half-baked, most notably since it was simultaneously dealing with item and player item changes.

As a result the final missing piece of the implementation, namely having metric events associated with the current item when obtained from the associated metric event publisher, was not working correctly. After playback end, for example, the list of metric events must be reset to an empty array, which was not the case.

I centered the `Tracker` again around a single `PlayerItem`, ensuring that `AVPlayerItem` updates are correctly applied. By separating metric event handling into a dedicated type the implementation can be made much simpler and tighter. The result can be seen on `944/per-item-tracking`, on which I also slightly polished a few things. This branch can be discussed next week.

defagos (approved):

PR #971, opened 2024-08-13
Add mandatory tracker support

defagos:
# Description

This PR adds support for mandatory trackers. When creating an associated adapter with mandatory behavior, the tracker will never be affected by the `isTrackingEnabled` property anymore. This is especially useful for QoS / QoE trackers that, unlike analytics trackers, might be useful in all cases.

# API design considerations

I considered many API design options and associated signatures, from a Boolean parameter to signatures including a _when_ plus some enum parameter, but none felt natural to read. The current proposal is the best I could come up with from an expressiveness point of view.

# Changes made

- Add new `TrackingBehavior` with `optional` and `mandatory` options.
- Add `behavior` parameter to tracker adapter creation methods. Default is `optional` (behavior until now).
- Fix Commanders Act tracking issue found when performing tests (a bug could prevent a tracker from being enabled in some scenario).
- Make an existing playlist-related test less flaky.

# Checklist

### DIFF
--- a/Sources/Player/Tracking/PlayerItemTracker.swift
+++ b/Sources/Player/Tracking/PlayerItemTracker.swift
@@ -97,8 +97,8 @@ public extension PlayerItemTracker where Configuration == Void {
     /// Creates an adapter for the receiver.
     /// 
     /// - Parameters:
-    ///   - mapper: A closure that maps an item metadata to tracker metadata.
     ///   - behavior: The tracking behavior.
+    ///   - mapper: A closure that maps an item metadata to tracker metadata.
     /// - Returns: The tracker adapter.
     static func adapter<M>(behavior: TrackingBehavior = .optional, mapper: @escaping (M) -> Metadata) -> TrackerAdapter<M> {
         .init(trackerType: Self.self, configuration: (), behavior: behavior, mapper: mapper)
--- a/Sources/Player/Tracking/PlayerItemTracker.swift
+++ b/Sources/Player/Tracking/PlayerItemTracker.swift
@@ -81,38 +81,45 @@ public extension PlayerItemTracker {
     ///
     /// - Parameters:
     ///   - configuration: The tracker configuration.
+    ///   - mandatory: Set to `true` to prevent the tracker from being disabled with with ``Player/isTrackingEnabled``.
     ///   - mapper: A closure that maps an item metadata to tracker metadata.
     /// - Returns: The tracker adapter.
-    static func adapter<M>(configuration: Configuration, mapper: @escaping (M) -> Metadata) -> TrackerAdapter<M> {
-        .init(trackerType: Self.self, configuration: configuration, mapper: mapper)
+    static func adapter<M>(configuration: Configuration, mandatory: Bool = false, mapper: @escaping (M) -> Metadata) -> TrackerAdapter<M> {
+        .init(trackerType: Self.self, mandatory: mandatory, configuration: configuration, mapper: mapper)
     }
 }
 
 public extension PlayerItemTracker where Configuration == Void {
     /// Creates an adapter for the receiver.
     /// 
-    /// - Parameter mapper: A closure that maps an item metadata to tracker metadata.
+    /// - Parameters:
+    ///   - mapper: A closure that maps an item metadata to tracker metadata.
+    ///   - mandatory: Set to `true` to prevent the tracker from being disabled with with ``Player/isTrackingEnabled``.
     /// - Returns: The tracker adapter.
-    static func adapter<M>(mapper: @escaping (M) -> Metadata) -> TrackerAdapter<M> {
-        .init(trackerType: Self.self, configuration: (), mapper: mapper)
+    static func adapter<M>(mapper: @escaping (M) -> Metadata, mandatory: Bool = false) -> TrackerAdapter<M> {
+        .init(trackerType: Self.self, mandatory: mandatory, configuration: (), mapper: mapper)
     }
 }
 
 public extension PlayerItemTracker where Metadata == Void {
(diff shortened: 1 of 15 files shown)


Sources/Player/Types/TrackingBehavior.swift
```diff
@@ -0,0 +1,20 @@
...
+    /// Optional tracking.
+    ///
+    /// The ``TrackerAdapter`` takes into account the ``Player/isTrackingEnabled`` player setting.
+    case optional
+
+    /// Mandatory tracking.
+    ///
+    /// The ``TrackerAdapter`` ignores the ``Player/isTrackingEnabled`` player setting.
+    case mandatory
+}
```
waliid, line 20: Isn't a boolean sufficient? (e.g. `isTrackingMandatory` or `isTrackingOptional`)

defagos: This was my first iteration but it felt clunky IMHO. 

Moreover, and though I did not implement it, we could have additional behaviors like "tracking if the app is in the foreground only". This is something we discussed in the context of our analytics requirements and an enum can be easily extended with additional behaviors, unlike a Bool.

TL;DR I found it both more expressive and extensible, thus this proposal.

defagos: Let me elaborate a bit more about the _clunkiness_ I was mentioning.

Booleans are great at describing simple self-describing binary states like `enabled` or `visible`, for example. They are less so at describing richer behaviors that cannot be described in simple terms, like `immediate` for our seeking behavior or the tracking behavior we are discussing. The tracking behavior is here not a simple state that can be understood in itself, it is merely a characterization of the interaction between a tracker behavior and how the parent player has been configured. It is only in response to this configuration that a tracker exhibits a certain behavior, it cannot be understood in isolation.

If you look at my first commit you can maybe see why this is clunky when a simple `mandatory` Boolean is used / read in code afterwards. If the Boolean is used as is (the `isTrackingMandatory` is probably too verbose, `mandatory` is the concept here since a tracker's purpose is to track), then you would have to add it everywhere this context is missing, because just seeing a _mandatory_ Boolean does not tell you what is mandatory (is it the thing you are setting? Is it another mandatory behavior like a user consent for example?). Thus each call site would likely need to be adorned with `isTrackingMandatory` or `isUserConsentMandatory` to provide for the missing context.

It is similar with simple words like `enabled`. While they can be great at conveying _light bulb_ states, they do not tell you which of these states is actually controlled. Once several states can have these on / off states, you need to provide context when they are mixed, otherwise you don't know which one is which.

waliid: I see your point, thanks for the explanation.

Sources/Player/Tracking/PlayerItemTracker.swift
```diff
@@ -82,37 +82,48 @@ public extension PlayerItemTracker {
     /// - Parameters:
     ///   - configuration: The tracker configuration.
     ///   - mapper: A closure that maps an item metadata to tracker metadata.
+    ///   - behavior: The tracking behavior.
```
waliid, line 85:
```suggestion
    ///   - behavior: The tracking behavior.
    ///   - mapper: A closure that maps an item metadata to tracker metadata.
```

Sources/Player/Tracking/PlayerItemTracker.swift
```diff
@@ -82,37 +82,48 @@ public extension PlayerItemTracker {
...
-    /// - Parameter mapper: A closure that maps an item metadata to tracker metadata.
+    /// - Parameters:
+    ///   - mapper: A closure that maps an item metadata to tracker metadata.
+    ///   - behavior: The tracking behavior.
     /// - Returns: The tracker adapter.
-    static func adapter<M>(mapper: @escaping (M) -> Metadata) -> TrackerAdapter<M> {
-        .init(trackerType: Self.self, configuration: (), mapper: mapper)
+    static func adapter<M>(mapper: @escaping (M) -> Metadata, behavior: TrackingBehavior = .optional) -> TrackerAdapter<M> {
+        .init(trackerType: Self.self, configuration: (), behavior: behavior, mapper: mapper)
     }
```
waliid, line 105:
```suggestion
    static func adapter<M>(behavior: TrackingBehavior = .optional, mapper: @escaping (M) -> Metadata) -> TrackerAdapter<M> {
        .init(trackerType: Self.self, configuration: (), behavior: behavior, mapper: mapper)
    }
```

Sources/Player/Tracking/PlayerItemTracker.swift
```diff
@@ -82,37 +82,48 @@ public extension PlayerItemTracker {
...
 }
 
 public extension PlayerItemTracker where Configuration == Void {
     /// Creates an adapter for the receiver.
     /// 
-    /// - Parameter mapper: A closure that maps an item metadata to tracker metadata.
+    /// - Parameters:
+    ///   - mapper: A closure that maps an item metadata to tracker metadata.
+    ///   - behavior: The tracking behavior.
```
waliid, line 101:
```suggestion
    /// - Parameters:
    ///   - behavior: The tracking behavior.
    ///   - mapper: A closure that maps an item metadata to tracker metadata.
```

waliid (approved):

waliid:
Great job, thank you for this PR!

Sources/Player/Tracking/PlayerItemTracker.swift
defagos: Fix parameter order

defagos:
Thanks for the review and for noticing parameter odering inconsistencies. Should be fixed now.

waliid (approved):

PR #1061, opened 2024-11-07
Fix player observability in response to media accessibility updates

defagos:
## Description

This PR fixes player observability in response to media accessibility updates. Event though we were observing such changes internally we were filtering identical values resulting from changes, preventing core property changes from being correctly published by the player in response to media accessibilty updates.

## Changes made

- Player properties (and associated objects) need to stay equatable so that duplicates can be easily removed and associated value streams can be easily sliced with `slice(at:)`. Since we cannot grab media accessibility settings as a whole when a `kMACaptionAppearanceSettingsChangedNotification` is received, I propose to publish the last update date instead (another option would have been a `UUID` but it felt less natural IMHO).
- `MediaSelectionProperties` now store the date of the most recent change, which is taken into account when checking for equality. As a result changes made to media accessibility changes now result in an externally observable core property change.
- Circumspect has been updated so that value output and observable change mutation can be observed with a nice formalism.
- Remove `testFailingUnauthorizedItemBetweenPlayableItems` flaky test. We had a lot of troubles with it recently and it was making this PR difficult to check for no good reason. I tried other approaches which seemed promising (e.g. increasing the length of the 1st stream involved) but in the end the test remained flaky.

## Checklist

### DIFF
--- a/Tests/CircumspectTests/Expectations/ExpectValueTests.swift
+++ b/Tests/CircumspectTests/Expectations/ExpectValueTests.swift
@@ -13,13 +13,13 @@ private class Object: ObservableObject {
     @Published var value = 0
 }
 
-final class ExpectOutputTests: XCTestCase {
-    func testSingleOutput() {
-        expectOutput(from: Just([1, 2, 3].publisher))
+final class ExpectValueTests: XCTestCase {
+    func testSingleValue() {
+        expectValue(from: Just([1, 2, 3].publisher))
     }
 
-    func testMultipleOutputs() {
-        expectOutput(from: Just([1, 2, 3].publisher))
+    func testMultipleValues() {
+        expectValue(from: Just([1, 2, 3].publisher))
     }
 
     func testSingleChange() {
--- a/Tests/CircumspectTests/Expectations/ExpectValueTests.swift
+++ b/Tests/CircumspectTests/Expectations/ExpectValueTests.swift
@@ -15,7 +15,7 @@ private class Object: ObservableObject {
 
 final class ExpectValueTests: XCTestCase {
     func testSingleValue() {
-        expectValue(from: Just([1, 2, 3].publisher))
+        expectValue(from: Just(1))
     }
 
     func testMultipleValues() {
--- a/Tests/CircumspectTests/Expectations/ExpectValueTests.swift
+++ b/Tests/CircumspectTests/Expectations/ExpectValueTests.swift
@@ -19,7 +19,7 @@ final class ExpectValueTests: XCTestCase {
     }
 
     func testMultipleValues() {
-        expectValue(from: Just([1, 2, 3].publisher))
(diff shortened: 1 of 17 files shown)


Tests/CircumspectTests/Expectations/ExpectValueTests.swift
```diff
@@ -0,0 +1,40 @@
...
+}
+
+final class ExpectValueTests: XCTestCase {
+    func testSingleValue() {
+        expectValue(from: Just([1, 2, 3].publisher))
+    }
+
+    func testMultipleValues() {
+        expectValue(from: Just([1, 2, 3].publisher))
+    }
```
waliid, line 23: This test seems to check exactly the same thing as the test above (`testSingleValue`). We could try using a sequence here instead of `Just` to handle several emissions. 

Here is my suggestion:

```suggestion
    func testMultipleValues() {
        expectValue(from: Publishers.Sequence<[Int], Never>(sequence: [1, 2, 3]))
    }
```

Tests/CircumspectTests/Expectations/ExpectValueTests.swift
```diff
@@ -0,0 +1,40 @@
...
+import XCTest
+
+private class Object: ObservableObject {
+    @Published var value = 0
+}
+
+final class ExpectValueTests: XCTestCase {
+    func testSingleValue() {
+        expectValue(from: Just([1, 2, 3].publisher))
+    }
```
waliid, line 19:
```suggestion
    func testSingleValue() {
        expectValue(from: Just(1))
    }
```

Tests/CircumspectTests/Expectations/ExpectValueTests.swift
```diff
@@ -0,0 +1,40 @@
...
+
+    func testMultipleChanges() {
+        let object = Object()
+        expectChange(from: object) {
+            object.value = 1
+            object.value = 2
+            object.value = 3
+        }
+    }
```
waliid, line 39: Testing with one value isn't enough?

waliid (approved):

Tests/CircumspectTests/Expectations/ExpectValueTests.swift
defagos: I don't know in which state I wrote this code 😅

There is a simpler alternative which requires less generics:

```swift
    func testMultipleValues() {
        expectValue(from: [1, 2, 3].publisher)
    }
```

This is what I pushed as a fix.

Tests/CircumspectTests/Expectations/ExpectValueTests.swift
defagos: That's the purpose here. Batching several observable object value changes of the same property. But maybe this test is not that relevant after all.

waliid (approved):

PR #1062, opened 2024-11-08
Improve demo accessibility via VoiceOver

defagos:
## Description

This PR makes various improvements to VoiceOver support in the demo.

## Changes made

- Improve custom player accessibility.
- Improve specific player accessibility.
- Implement Z-escape gesture for modals.
- Make metrics debugging view accessible.
- Make all lists accessible.
- Make settings accessible.

It is likely that all accessibility strings can be properly localized but this should be a pretty good start already. The tvOS implementation might not be perfect as well but could likely be enhanced in a second PR.

## Checklist

### DIFF
--- a/Demo/Sources/Players/PlaybackView.swift
+++ b/Demo/Sources/Players/PlaybackView.swift
@@ -508,6 +508,7 @@ private struct LoadingIndicator: View {
             .tint(.white)
             .opacity(isBuffering ? 1 : 0)
             .animation(.linear(duration: 0.1), value: isBuffering)
+            .accessibilityHidden(true)
             .onReceive(player: player, assign: \.isBuffering, to: $isBuffering)
     }
 }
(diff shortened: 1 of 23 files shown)


Demo/Sources/Players/PlaybackView.swift
```diff
@@ -108,12 +108,16 @@ private struct MainView: View {
     private func main() -> some View {
         ZStack {
             video()
+                .accessibilityElement()
+                .accessibilityLabel("Video")
+                .accessibilityHint("Double tap to toggle controls")
+                .accessibilityAction(.default, visibilityTracker.toggle)
```
waliid, line 114: Instead of having this hint and this action, should we keep controls always visible when **VoiceOver** is on?

waliid (approved):

Demo/Sources/Players/PlaybackView.swift
defagos: The fact that controls are persistent but togglable when VoiceOver is on is a desired feature. We cannot assume that only blind users use VoiceOver. People who still can see must be able to use the player, and they might want to have controls off so that they can still see the image. This is also how Apple standard player UI behaves.

PR #1104, opened 2024-12-13
Cleanup Makefile targets

waliid:
## Description

This PR simplifies our Makefile targets by using dedicated scripts for each one.

## Changes made

- Scripts have been split in two directories, `public` and `private`.
    - Scripts located in `public` directory are use by our **Makefile**.
    - Scripts located in `private` directory are use by our GitHub workflows.

## Checklist


Scripts/test-streams.sh
```diff
@@ -3,6 +3,7 @@
 SCRIPT_NAME=$(basename "$0")
 SCRIPT_DIR=$(dirname "$0")
 
+Scripts/install-pkgx.sh
```
defagos, line 6: This prevents the script from being run from any directory. We can assume that `install-pkgx.sh` is in the same directory, thus maybe something like

```suggestion
"$(dirname -- "$0")"/install-pkgx.sh
```

(there might be a better way to write it) would make it possible to run the script from everywhere.

Scripts/set-apple-api-env-interactively.sh
```diff
@@ -0,0 +1,28 @@
...
+
+if [[ -z "$TEAM_ID" || -z "$KEY_ISSUER_ID" || -z "$KEY_ID" || -z "$APPLE_API_KEY_BASE64" ]]; then
+    echo
+    [[ -z "$TEAM_ID" ]] && echo "  TEAM_ID is missing!"
+    [[ -z "$KEY_ISSUER_ID" ]] && echo "  KEY_ISSUER_ID is missing!"
+    [[ -z "$KEY_ID" ]] && echo "  KEY_ID is missing!"
+    [[ -z "$APPLE_API_KEY_BASE64" ]] && echo "  APPLE_API_KEY_BASE64 is missing!"
+    exit 1
+fi
+echo
```
defagos, line 23: Wouldn't it be simpler to perform a sanity check after each `read`?

Scripts/test.sh
```diff
@@ -0,0 +1,30 @@
...
+if [[ "$1" != "--platform" ]]; then
+    usage
+fi
+
+PLATFORM="$2"
+if [[ $PLATFORM != "ios" && $PLATFORM != "tvos" ]]; then
+    usage
+fi
+
+Scripts/test-streams.sh -s
```
defagos, line 21: I would maybe keep test stream management and test execution separate (only bundled together in the makefile), only combining them at the Makefile level.

Scripts/find-dead-code.sh
defagos, line 1: `pkgx` might not be installed. Shouldn't we install as in other scripts?

Scripts/clean-imports.sh
defagos, line 1: pkgx might not be installed. Shouldn't we install as in other scripts?

defagos (commented):
Two general comments which might already appear assigned to some code changes:

- Parameter management is likely better with `getopts`. Several parameters have been introduced an I think having a way to reliably manage them would be beneficial.
- I like the idea of having extracted shell scripts for everything. I am still a bit wondering if we shouldn't make each script standalone. Some scripts have been introduced to factor out common logic (e.g. interactively getting secrets, installing pkgx, installing bundler) but shell scripts are difficult to read, understand and maintain when scattered over several files, each possibly defining its own set of variables used in other files.

Makefile
```diff
@@ -3,167 +3,105 @@
...
+	@echo
+	@echo "Archive:"
+	@echo "  archive-demo-ios               Archive iOS demo [MODE=\"--non-interactive\"]"
+	@echo "  archive-demo-tvos              Archive tvOS demo [MODE=\"--non-interactive\"]"
+	@echo
+	@echo "Deliver:"
+	@echo "  deliver-demo-nightly-ios       Deliver nightly iOS demo build [MODE=\"--non-interactive\"]"
+	@echo "  deliver-demo-nightly-tvos      Deliver nightly tvOS demo build [MODE=\"--non-interactive\"]"
+	@echo "  deliver-demo-release-ios       Deliver release iOS demo build [MODE=\"--non-interactive\"]"
+	@echo "  deliver-demo-release-tvos      Deliver release tvOS demo build [MODE=\"--non-interactive\"]"
```
defagos, line 86: Non-interactive mode is mostly for internal use. Not sure it should be visible in documentation.

defagos (approved):

defagos (approved):

PR #1184, opened 2025-04-01
Improve playlist demo item insertion

defagos:
## Description

This PR improves playlist demo item insertion support with inspiration from our recent [Castor demo implementation](https://github.com/SRGSSR/castor/pull/61).

## Changes made

- Update model to support insertion of the same media several times.
- Conform `PlayerItem` hashable to make it more suitable for wrapping into structs which themselves need to be hashable. Instead of approaches we previously used (generic source attached to the item (#1041), sync between media and player item lists we had until this PR), this makes it possible to use composition as a simple way of implementing playlist items that bundle a business object and its associated item.
- Add insertion location options (prepend, before current item, after current item, append).
- Add multiplier to insert the same group of items several times.
- Better organize playlist implementation and mark iOS-specific list management as unavailable on tvOS.

## Checklist

### DIFF
--- a/Demo/Sources/Showcase/Playlist/PlaylistSelectionView.swift
+++ b/Demo/Sources/Showcase/Playlist/PlaylistSelectionView.swift
@@ -6,6 +6,7 @@
 
 import SwiftUI
 
+@available(tvOS, unavailable)
 struct PlaylistSelectionView: View {
     static let medias = [
         URLMedia.onDemandVideoHLS,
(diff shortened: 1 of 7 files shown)


Demo/Sources/Showcase/Playlist/PlaylistSelectionView.swift
```diff
@@ -67,12 +71,70 @@ struct PlaylistSelectionView: View {
...
+        .padding()
+    }
+
     private func cancel() {
         dismiss()
     }
 
     private func add() {
-        model.add(Array(selectedMedias))
+        let entries = Array(repeating: selectedMedias, count: multiplier).flatMap(\.self).map { PlaylistEntry(media: $0) }
```
waliid, line 104:
```suggestion
        let entries = Array(repeating: selectedMedias, count: multiplier).flatMap(\.self).map(PlaylistEntry.init)
```

waliid (approved):

waliid (approved):

PR #1193, opened 2025-04-10
Update macOS and Xcode versions

waliid:
## Description

This PR updates the version of Xcode used to 16.3 and also set the macos version used for the CI.

## Changes made

Self-explanatory.

## Checklist


defagos (approved):

defagos:
There has been a change in how `AVQueuePlayer` consumes failing items in 18.4. This can be observed by running the following sample code with 18.4 and an earlier version:

```swift
import AVFoundation
import AVKit
import Combine
import SwiftUI

struct ContentView: View {
    @State private var player = AVQueuePlayer(items: [
        // Before and after 18.4: Consumed
        // .init(url: URL(string: "http://localhost:8123/unavailable/master.m3u8")!),

        // Before 18.4: Stuck on failed item. After 18.4: Consumed
        // .init(url: URL(string: "https://httpbin.org/status/403")!),

        // Before 18.4: Stuck on failed item. After 18.4: Consumed
        .init(url: URL(string: "http://localhost:8123/unavailable.mp3")!),

        .init(url: URL(string: "https://devstreaming-cdn.apple.com/videos/streaming/examples/bipbop_16x9/bipbop_16x9_variant.m3u8")!)
    ])

    var body: some View {
        VideoPlayer(player: player)
            .onAppear(perform: player.play)
            .ignoresSafeArea()
    }
}

#Preview {
    ContentView()
}
```

Before iOS/tvOS 18.4 the player was not consuming failed items in some cases, which we considered the exception rather than the rule. In iOS/tvOS 18.4 this behavior appears to be the rule. This is actually an improvement since the behavior is now consistent: `AVQueuePlayer` consumes items after finishing playback, whether successfully or because of a failure.

Even though our player queue sync implementation does not rely on this behavior (we avoid filling the queue to avoid automatic consumption of failed items), our UTs were checking the internal `AVQueuePlayer` behavior to find such changes in behavior. We will therefore keep these internal behavior checks but adjust them to accommodate for fixes in 18.4.

PR #1200, opened 2025-04-23
Allow to start Picture-in-Picture automatically for inline contents

waliid:
## Description

Self-explanatory.

## Changes made

- The property `canStartPictureInPictureAutomaticallyFromInline` is to `true` by default.

## Checklist


defagos (approved):

defagos:
Note that we could have made this behavior configurable on `PictureInPicture` directly, but in general the behavior obtained by allowing PiP to start from inline is far more consistent and understandable from a user perspective, so we decided to simply set it to `true`.

PR #1267, opened 2025-08-11
Support transient resume item and time

defagos:
## Description

This PR adds an API to set an item and position to resume at.

## API design considerations

Three API naming possibilities have been considered:

- `Player.resume(_:in:)`
- `Player.seek(_:in:)`
- `Player.setCurrentItem(_:position:)`

The third option did not feel very natural and was dropped early. The second option was the obvious choice at first but the API was not consistent with other seek APIs we provide, which all offer a completion handler. For this reason a new API concept was introduced so that there is no confusion between these APIs.

## Implementation considerations

Since we want the API to be usable before the player even begins to play content we need to store the position. But unlike `PlaybackConfiguration` which applies a start position consistently each time an item is played, our new API must have transient behavior, i.e. subsequent playback attempts must not be affected (except of course if the resume API is used again in between playbacks).

For this reason we need some storage which associates the item to be affected and the position that must be applied. This storage must be able to self-erase its content when it has been read by the matching item.

## Known limitations

If playback fails with a valid `AVPlayerItem` (e.g. URL-based item or during playback), replaying the content will start at the default position. This means that currently:

- If a URL-based item is played with no network is available, then replayed once the network is available again, the initial resume position is lost.
- If a custom item (e.g. URN-based) is played with no network is available, then replayed once the network is available again, the initial resume position is applied.

## Changes made

- Add `Player.resume(_:in:)` method.
- Revisit the implementation to avoid superfluous dependencies on `PlayerLimits`.
- Add self-destructing `ResumeState` type to store a resume position associated with an id. Reading a state is made through an id to avoid the state be applied to items it is not associated with.
- Update `PlaybackConfiguration` item configuration method to take a resume state as additional parameter. This is required since this method ensures that per-item start positions are correctly applied while ensuring blocked time ranges cannot be reached.

## Checklist


Sources/Player/Types/ResumeState.swift
```diff
@@ -0,0 +1,29 @@
...
+
+import Foundation
+
+final class ResumeState {
+    private var position: Position?
+    private let id: UUID
+
+    var isEmpty: Bool {
+        position == nil
+    }
```
waliid, line 15: Do we really need this computed property? It seems to be used only for a testing purposes.

waliid (approved):

waliid (approved):

PR #1277, opened 2025-08-22
Support empty media selection

defagos:
## Description

When [returning from Google Cast playback](https://github.com/SRGSSR/castor/issues/144) in our Castor demo implementation, the user might have disabled subtitles on the receiver. In such cases it is natural to attempt to restore the same setting on a Pillarbox instance (used in the Castor demo as local player) when resuming playback locally, but `Player.setMediaSelection(preferredLanguages:for:)` currently provides no way to prefer no selection at all before starting playback, since setting an empty list removes media selection overrides instead.

This PR enhances our API to support setting no language at all as media selection preference (if possible; some characteristics might force a selection).

## Changes made

- Introduce `MediaSelectionPreference` to convey media selection preference settings. This type is a `struct` so that languages can be specified as a variadic parameter (thus forcing at least one value to be provided and enums do not support variadic associated values). Values `.automatic`, `.on` and `.off` are provided, mirroring the usual media options.
- Update media selection API to take a `MediaSelectionPreference` as parameter (breaking API change).
- Remove API to read media selection preference, an operation with no real added value.
- Publish a `Player` update when the media selection changes. Previously we could change the media selection but the player would not publish any observable change, preventing current option selection from being reflected until the next update. Since media selection is part of core player updates it is natural that media selection changes must lead to similar updates as well.

## Checklist

### DIFF
--- a/Sources/Player/Player/Player+MediaSelection.swift
+++ b/Sources/Player/Player/Player+MediaSelection.swift
@@ -113,7 +113,7 @@ public extension Player {
     ///     include `.audible`, `.legible`, and `.visual`.
     ///
     /// This method can be used to override the default media option selection for some characteristic, e.g., to start
-    /// playback with a predefined language for audio and / or subtitles.
+    /// playback with a predefined language for audio and/or subtitles.
     ///
     /// > Important: Media selection only works when HLS playlists are correctly formatted. If selection does not behave
     ///   as expected, see the troubleshooting section in <doc:stream-encoding-and-packaging-advice-article> to identify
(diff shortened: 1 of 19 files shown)


defagos:
There is a bug. Restoring the last selection from MediaAccessibility does not work anymore:

1. Play Apple basic 16:9.
2. Change subtitles to Japanese.
3. Reopen the same content. Subtitles are not in Japanese.

Sources/Player/Player/Player+MediaSelection.swift
```diff
@@ -133,10 +143,17 @@ public extension Player {
...
+            queuePlayer.setMediaSelectionCriteria(Self.emptyMediaSelectionCriteria(for: characteristic), forMediaCharacteristic: characteristic)
         }
     }
 
+    /// Resets media selection preferred languages for the specified media characteristic.
+    ///
+    /// This method removes any overrides, returning the behavior to default media option selection.
+    func resetMediaSelectionPreferredLanguages(for characteristic: AVMediaCharacteristic) {
+        queuePlayer.setMediaSelectionCriteria(nil, forMediaCharacteristic: characteristic)
+    }
```
waliid, line 155: What do you think about keeping only one method `setMediaSelection` with an enum  `PreferredLanguages` (`.none`, `.codes([...])`, `.default`) instead of having two methods (`set` and `reset`)?

waliid: Seen together.

PR #1311, opened 2025-10-03
Clean up Control Center on error

waliid:
## Description

This PR ensures that the **Control Center** is properly cleaned up when an error occurs.

## Changes made

- `nowPlayingPublisher()` has been updated to take into account errors.

| Issue | Fix |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/590ff415-89c9-4d98-993b-e879c2e3fd2e" /> | <video src="https://github.com/user-attachments/assets/83598684-0527-4887-bac1-e371f7545aee" /> |

## Checklist


Sources/Player/Player/Player+ControlCenter.swift
```diff
@@ -127,9 +127,9 @@ extension Player {
     }
 
     func nowPlayingPublisher() -> AnyPublisher<NowPlaying, Never> {
-        Publishers.CombineLatest(isActivePublisher, queuePublisher)
-            .map { [weak self] isActive, queue in
-                guard let self, isActive, !queue.isActive else { return Just(NowPlaying.empty).eraseToAnyPublisher() }
+        Publishers.CombineLatest3(isActivePublisher, queuePublisher, queueErrorPublisher())
+            .map { [weak self] isActive, queue, error in
+                guard let self, isActive, !queue.isActive, error == nil else { return Just(NowPlaying.empty).eraseToAnyPublisher() }
```
defagos, line 132: The queue publisher already reacts to errors, so we should simply test `queue.error` in the map operator closure here. I checked this intuition and the test you added passes.

```suggestion
                guard let self, isActive, !queue.isActive, queue.error == nil else { return Just(NowPlaying.empty).eraseToAnyPublisher() }
```

If this matches your expectation as well I guess we can then revert the other changes since they are not required.

PR #1368, opened 2025-11-17
Mitigate slow player deallocation on tvOS

defagos:
## Description

This PR mitigates slow player deallocation observed on tvOS, seemingly related to the use of long-running asset resource loader delegates in our implementation.

For more information about this implementation, please read the associated issue.

## Changes made

- Retain loading/renewal requests in resource loader delegate implementation.
- Avoid reusing the same resource loader delegate for all items inserted during metadata retrieval.
- Cancel asset loading for an item when replaced by another one (must wait a little bit to avoid issues).
- Refactor `AVQueuePlayer.replaceItems(with:)` to isolate related workarounds.
- Update Nimble to version 13.8.0 (should fix potential main thread hangs).
- Remove volume set to 0 when `Player` is deallocated (not needed anymore with the fix introduced in this PR).

## Checklist

### DIFF
--- a/Sources/Player/Extensions/AVQueuePlayer.swift
+++ b/Sources/Player/Extensions/AVQueuePlayer.swift
@@ -11,7 +11,6 @@ extension AVQueuePlayer {
         guard self.items() != items else { return }
 
         if let firstItem = items.first {
-            let currentItem = self.items().first
             if firstItem !== currentItem {
                 safelyReplaceCurrentItem(with: firstItem)
             }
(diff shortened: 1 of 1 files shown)


Sources/Player/Extensions/AVQueuePlayer.swift
```diff
@@ -39,4 +34,25 @@ extension AVQueuePlayer {
     private func append(_ items: [AVPlayerItem]) {
         items.forEach { insert($0, after: nil) }
     }
+
+    private func safelyReplaceCurrentItem(with item: AVPlayerItem) {
+        // TODO: Workaround to fix incorrect recovery from failed MP3 playback (FB13650115). Remove when fixed.
+        if let currentItem = items().first, currentItem.error != nil {
```
waliid:
```suggestion
        if let currentItem, currentItem.error != nil {
```

PR #1444, opened 2026-02-17
Improve standard connector documentation

waliid:
## Description

Self-explanatory.

## Checklist

### DIFF
--- a/Sources/StandardConnector/PlayerData.swift
+++ b/Sources/StandardConnector/PlayerData.swift
@@ -10,8 +10,6 @@ import PillarboxPlayer
 
 // FIXME: Remove when minimum target is 17 and use `Never` instead.
 /// An object representing the absence of custom data.
-///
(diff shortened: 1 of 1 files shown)


Sources/StandardConnector/PlayerData.swift
```diff
@@ -10,9 +10,14 @@ import PillarboxPlayer
 
 // FIXME: Remove when minimum target is 17 and use `Never` instead.
 /// An object representing the absence of custom data.
+///
+/// Use `Never` when targeting iOS 17 or later.
```
defagos: it is likely better to have clients use `EmptyCustomData` until the minimum target is iOS 17. Usage will likely be transparent in client code (thanks to the prototype with empty custom data) and this avoids painful conditional code:

```suggestion
```

Once the minimum target is iOS 17 we will be able to document as breaking change that `EmptyCustomData` has been replaced with `Never`.

PR #1462, opened 2026-03-21
Define asset loading logic through types

defagos:
## Description

This PR wraps asset loading logic into dedicated types conforming to the new `AssetLoader` protocol. This will allow us to persist what kind of asset loader is used by an item, making it possible to serialize/deserialize the asset loading logic associated with a download. This will allow us to:

- Resume downloads, even if metadata wasn't loaded yet.
- Restart failed downloads.
- Fetch images that might have failed to download again.

This PR introduces several breaking changes but most of them should be pretty easy to address. Only minor changes were required in the demo, which proves most developers should have no or minimal work related to this change.

## Changes made

- Add `AssetLoader`, defining which input is required to load content, which metadata is associated with it, how the content is fetched and how it must be mapped to `PlayerMetadata`. This avoids requiring conformance to a protocol for defining player metadata, which felt a bit intrusive.
- Update `PlayerItem` construction to support custom asset loader types instead of a publisher.  Where no metadata is required `Void` was replaced with `PlayerMetadata.empty` instead (with `.empty` now publicly exposed).
- Remove `AssetMetadata` protocol. Instead of having to conform to this protocol, use constructors with a `PlayerMetadata` and build the metadata directly at call site (possibly using a helper method, an approach which was applied in this PR).
- Update StandardConnector and CoreBusiness packages to use dedicated asset loaders.
- Update `PlayerItemTracker` and adapters to support `PlayerMetadata` natively, with `.empty` replacing `Void` when no metadata is provided.
- Update documentation and tutorials.
- Update the demo.
- Fix Pillarbox standard connector scheme.

### Remark

If asset loader input/metadata need to be serialized later we can simply add protocol methods to `AssetLoader`, which take an input or metadata as parameter and must return some codable type (instead of asking for `Input` and `Metadata` to themselves conform to `Codable`, which is more involved).

## Checklist


Sources/Player/Player.docc/Articles/playback/playback-article.md
```diff
@@ -58,15 +58,21 @@ You can initialize a player with or without media items. Since ``Player`` confor
 
 ## Configure the player
 
-Customize the player during initialization by providing a ``PlayerConfiguration`` object. Note that configuration settings are immutable after the player is created.
+Customize the player during initialization by providing a ``PlayerConfiguration`` object. Note that configuration settings are immutable after the player has been created.
 
 ## Load custom content
 
-In scenarios where media URLs are dynamically retrieved (e.g., from a web service), you can create custom player items by following these steps:
+When your content is retrieved dynamically—such as from a web service—you can create custom player items by implementing an asset loader tailored to your needs.
```
waliid:
```suggestion
When your content is retrieved dynamically, such as from a web service, you can create custom player items by implementing an asset loader tailored to your needs.
```

waliid (approved):

PR #1471, opened 2026-04-09
Avoid measuring time while the device is asleep

### LINKED ISSUE #1469: Time measurement issues
**As a quality engineer** I need to be able to evaluate the QoE and QoS correctly. Currently Apple players often return times that are too large.

Image

A list of sessions can be found by exploring our data with Grafana:

Image

## Hints

We conjecture that date-based management does not account for the app being potentially suspended, leading to potentially large date intervals. We should use `SuspendingClock` instead.

## Acceptance criteria

- Large QoS/QoE timings have been minimized.
- Once delivered to production, the Lucene query above, limited to fixed versions of Pillarbox, does provide values that are consistently smaller, e.g.

    ```
    session.player.platform:Apple AND session.qoe_timings.total:>500000 AND session.player.version:18.0.0
    ```

## Tasks

- [x] Fix measurement helpers to use a suspending clock rather than dates.
- [x] Update metrics to return a time interval instead of a date interval (dates are not really interesting here).
- [x] Fix `Stopwatch` implementation.

defagos:
## Description

This PR implements timing measurements (QoS/QoE) using a continuous clock instead of date comparisons. This should better account for periods during which the device is asleep, which were leading to large measured values.

### Remark

`SuspendingClock` does not seem to suspend while the device is plugged to an electrical outlet.

## Changes made

- Introduce clock interval that is similar to a `CMTimeRange`, but with clock instants and durations.
- Update interval measurement operator to deliver clock intervals.
- Introduce new interval measurement operator that delivers values and measures the interval between them at the same time.
- Update `Stopwatch` to use a clock.
- Update metrics gathering logic to use a continuous clock for measurements.
- Update trackers to use a continuous clock.
- Display device identifier in the demo settings view.
- Add sharing button in addition to string copy cell swipe actions (for more flexible transfer between devices not associated with the same Apple ID).

### Breaking changes

- APIs to measure time, part of Pillarbox Core, have been updated to return `Duration`/`Instant` information, grouped as `ClockInterval`, instead of `DateInterval`.
- Metadata/asset metric events are associated with durations instead of date intervals. Dates are not interesting anyway, the relevant data is how much time was needed.

## Checklist


Sources/Core/Extensions/Publisher.swift
```diff
@@ -186,18 +186,28 @@ public extension Publisher {
...
+    ///   consecutive elements.
+    func withInterval<C>(clock: C = .continuous) -> AnyPublisher<(Output, ClockInterval<C>), Failure> where C: Clock {
+        map { ($0, clock.now) }
+            .withPrevious((Optional<Output>.none, clock.now))
+            .compactMap { previous, current in
+                guard let value = current.0 else { return nil }
+                return (value, ClockInterval(start: previous.1, duration: previous.1.duration(to: current.1)))
             }
             .eraseToAnyPublisher()
     }
```
waliid:
```suggestion
    func withInterval<C>(clock: C = .continuous) -> AnyPublisher<(output: Output, interval: ClockInterval<C>), Failure> where C: Clock {
        map { ($0, clock.now) }
            .withPrevious((output: Optional<Output>.none, instant: clock.now))
            .compactMap { previous, current in
                guard let output = current.output else { return nil }
                return (output, ClockInterval(start: previous.instant, duration: previous.instant.duration(to: current.instant)))
            }
            .eraseToAnyPublisher()
    }
```

Sources/Core/Extensions/Publisher.swift
```diff
@@ -186,18 +186,28 @@ public extension Publisher {
...
+
+    /// Measures the time interval between consecutive outputs.
+    ///
+    /// - Parameter clock: The clock to use.
+    /// - Returns: A publisher that emits elements representing the time interval between the elements it receives.
+    func measureInterval<C>(clock: C = .continuous) -> AnyPublisher<ClockInterval<C>, Failure> where C: Clock {
+        withInterval(clock: clock)
+            .map(\.1)
```
waliid:
```suggestion
            .map(\.interval)
```

waliid (approved):

PR #1498, opened 2026-05-30
Support offline metadata persistence 

### LINKED ISSUE #1459: Offline player metadata management
**As a user** I want to be able to play a locally downloaded stream in the same conditions as when the content is streamed. This means:

- Integration with the Control Center (title, artwork).
- Chapter support.
- Time ranges (e.g. skipping over blocked intervals or credits).

## Acceptance criteria

- The API provides a way to retrieve content with associated standard player metadata.
- Standard player metadata is used locally to provide the same experience as the streaming experience, as described above. 

## Tasks

- [ ] Make `AssetContent` conform to `Codable` (`Resource` is a bit more challenging, but maybe a strategy similar to `AssetLoader` could work).
- [ ] Use `AssetLoader` as first step to fetch download metadata before the task can actually be started.

defagos:
## Description

This PR implements persistence for inputs and metadata associated with downloads.

The main challenge was to offer a flexible store concept that can fully preserve data associated with an asset, while integrating with Pillarbox player nicely. To achieve this goal the following behaviors were implemented:

- A protocol `AssetDownloadStore` has been introduced, with which applications can decide how they persist data associated with a download. By moving persistence to the application level we ensure that each app can fully decide how it manages data associated with downloads (including migration between versions of its data model).
- Inputs and metadata are both persisted for each download. It is therefore possible to reconstruct a download from the start even if the download immediately failed (e.g. no network connection during initial metadata retrieval) or is restarted at a later time (e.g. after the associated file was purged by the system).
- Persisted metadata can be used to build offline `PlayerMetadata`, which means all usual Pillarbox features are theoretically supported offline as well (metadata display, Control Center, chapters, blocked time ranges, credits). This PR does not offer a full implementation of all these features yet, though. Complete metadata and image support will be added in separate PRs.
- Arbitrary asset metadata can be persisted. Provided essential metadata they need has been persisted, trackers can therefore work online or offline. This makes features like offline analytics tracking possible (this of course requires trackers to have their own offline persistence mechanism).

Another challenge was to ensure that local download data stays consistent, most notably that no downloaded package (see #1456) is left dangling (orphan streams can namely waste a lot of space). We also wanted to be able to restart any download, should the system discard its associated local package to free up space, or should the user have deleted it from system storage settings.

A further challenge was to avoid breaking Pillarbox `Player` itself by introducing downloads too tightly with it. The design proposed in this PR achieves this result by fully reusing the native asset loading mechanism introduced in #1462 without requiring any change to `Player` itself. Downloads are therefore an entirely separate feature, seamlessly integrating with Pillarbox `Player`, and which could perfectly be moved to a dedicated package.

Finally, we also wanted to have something that can be automatically tested, especially since `AVAssetDownloadURLSession` itself does not work in test runners. By identifying which specific part of the download process it controls and abstracting it using a `DownloadSession`, we can simulate a local download in test runners as well, with great confidence things will work identically with `AVAssetDownloadURLSession`. Each test we have was run individually 100 times in sequence to iron out as many potential issues as possible with our implementation.

Ready-to-use store implementations are still possible, though, especially for our Standard Connector and Core Business packages. Those will be provided in separate PRs.

## Changes made

- Update `AssetLoader` to separate metadata retrieval from asset extraction. An asset which an be played online might namely differ from the corresponding downloadable asset (e.g. different stream URL, asset resource loader delegate not available when downloading files). By default a downloadable asset is assumed to be identical to the online asset, but the `AssetLoader` protocol provides optional customization should this not be the case. Note that the protocol method signature will likely evolve to return an asset publisher, as the process of building a downloadable asset might be asynchronous (e.g. token retrieval once before download).
- Update `AssetLoader` player metadata construction method to support input as parameter. Usually the input associated with a content to be played might namely be obscure (e.g. just an identifier). For downloads, or even for playback, we might want to supply minimal input information when creating the item or the download. This way minimal user-friendly information can be built and associated with an item or a download.
- Add `AssetDownloadStore` protocol which defines how input and metadata associated with a download, bundled as a `DownloadRecord`, are inserted, updated, read or removed. To ensure a download can be uniquely identified, a store static method  is responsible to build a stable identifier from a download's input. This also makes it possible to check local downloads for existence/status anywhere in an app, e.g. when implementing a download button associated with a content.
- Add `DownloaderManager`, responsible of download management (creation, retrieval, removal). To avoid exposing too many types unnecessarily this class is intern, with a public `Downloader` erasing types as much as possible (except at construction time). The public class is an `ObservableObject` as well, which makes direct integration with SwiftUI code possible.
- Add `Download`, responsible of a single download life cycle, combining an initial metadata retrieval phase and a file download phase. Both phases are consolidated into a single life cycle which delivers published properties (state, progress, file URL) as a stream. Essential properties are exposed publicly and thus automatically observable in SwiftUI views.

## Design considerations

- As a general rule, when conversion methods are defined, they are not defined by forcing types to conform to protocols directly. For example identifier construction from input has not been implemented by requiring inputs to be `Identifiable`, but rather as a store static method. The definition of what identifies a download uniquely is namely specific and forcing a unique behavior through protocol conformance is too rigid, should different interpretations have to coexist.
- Downloads require asset loader and store types to be provided, with constraints between them. A store must namely be able to persist input and metadata matching those of a data source, otherwise we would have required additional two-way mappings to be provided each time a store is combined with an asset loader, which was not ergonomic. To avoid these types leaking outside the SDK, we erased types from `DownloadManager` and `Download` as much as possible, so that views can for example be built for any kind of download, without littering view types with generics
- A `Download` is quite similar to a `PlayerItem`. Both have no generics (though they are built starting with an asset loader type and associated input) and both implement their life cycle in a similar way. Similarly, `Downloader` plays the same role for a `Download` as `Player` for a `PlayerItem`. Playback and download types are separate, though, with `AssetLoader` ensuring that business logic can be shared between them consistently.

## Other changes

- Remove generic metadata from `Asset`.
- Hide Core Business raw player item creation helpers. Those have no associated SRG SSR tracking and should not be promoted. They are instead hidden for usage from our demo only. Since most of these non-public helpers were not using Commanders Act anyway the `PlaybackContext` type has been removed as well (it was inconsistently used and mostly useless in the end). 
- Implement simple file-based download persistence in the demo. This will be further improved in upcoming PRs, e.g. using SwiftData instead.
- Improve download list with status and error display. Examples are still hardcoded URLs but we will be able to persist SRG SSR content in upcoming PRs instead.
- Update test streams with a local MP3 file that can be downloaded.
- Remove unnecessary type constraints on expectation publishers, letting them be used with all kinds of publishers, not only with those which can never fail.
- Add Combine fail operator which can be used to make an upstream publisher fail with an error on-demand. This operator makes it possible to selectively embed failure sources that can be activated by a simple `Trigger`, e.g. to implement remote-triggered cancellation. Initial designs simply used `prefix(untilOutputFrom:)` but this operator lacked feedback downstream (we couldn't determine if a process was cancelled or not, which is important if we want to have correct status information).
- Revisit player item mocks to have a more consistent and versatile API.
- Update tests relying on required values to use Nimble unwrapping, ensuring failed conditions do not crash the test but are rather reported as expectation failures instead.
- Disable `prefer_self_in_static_references` linter rule (too many false positives with [SwiftLint 0.63.3](https://github.com/realm/SwiftLint/releases/tag/0.63.3)).

## Checklist


Demo/Sources/Downloads/DemoAssetDownloadStore~ios.swift
```diff
@@ -0,0 +1,113 @@
...
+
+        init(id: String, record: DownloadRecord<DemoAssetLoader.Input, String>) {
+            self.id = id
+            self.title = record.input.title
+            self.url = record.input.url
+            self.metadata = record.metadata
+            self.bookmarkData = record.bookmarkData
+            self.progress = record.progress
+            self.errorDescription = record.error?.localizedDescription
+        }
```
waliid:
```suggestion
private init(id: String, title: String, url: URL, metadata: String?, bookmarkData: Data?, progress: Double, errorDescription: String?) {
            self.id = id
            self.title = title
            self.url = url
            self.metadata = metadata
            self.bookmarkData = bookmarkData
            self.progress = progress
            self.errorDescription = errorDescription
        }

        init(id: String, input: DemoAssetLoader.Input) {
            self.init(
                id: id,
                title: input.title,
                url: input.url,
                metadata: nil,
                bookmarkData: nil,
                progress: 0,
                errorDescription: nil
            )
        }

        init(id: String, record: DownloadRecord<DemoAssetLoader.Input, String>) {
            self.init(
                id: id,
                title: record.input.title,
                url: record.input.url,
                metadata: record.metadata,
                bookmarkData: record.bookmarkData,
                progress: record.progress,
                errorDescription: record.error?.localizedDescription
            )
        }
```

Demo/Sources/Downloads/DemoAssetDownloadStore~ios.swift
```diff
@@ -0,0 +1,113 @@
...
+//
+
+import Foundation
+
+@_spi(DownloaderPrivate)
+import PillarboxPlayer
+
+#if DEBUG
+
+private struct DownloadError: LocalizedError {
```
waliid, line 14:
```suggestion

#if DEBUG

import Foundation

@_spi(DownloaderPrivate)
import PillarboxPlayer

private struct DownloadError: LocalizedError {
```

Demo/Sources/Downloads/DemoAssetLoader~ios.swift
```diff
@@ -0,0 +1,38 @@
...
+
+import Combine
+import Foundation
+
+@_spi(DownloaderPrivate)
+import PillarboxPlayer
+
+#if DEBUG
+
+struct DemoAssetLoader: AssetLoader {
```
waliid, line 15:
```suggestion

#if DEBUG

import Combine
import Foundation

@_spi(DownloaderPrivate)
import PillarboxPlayer

struct DemoAssetLoader: AssetLoader {
```

Demo/Sources/Downloads/DownloadCell~ios.swift
```diff
@@ -13,7 +13,12 @@ import SwiftUI
 
 struct DownloadCell: View {
     @ObservedObject var download: Download
-    @EnvironmentObject private var router: Router
+
+    var action: () -> Void
```
waliid:
```suggestion
    private(set) var action: () -> Void
```

Demo/Sources/Downloads/DownloadCell~ios.swift
```diff
@@ -24,54 +29,62 @@ struct DownloadCell: View {
...
-                router.presented = .player(media: .init(title: download.title, type: .url(url)))
-            case .failed:
-                download.restart()
-            default:
-                break
-            }
+        .onTapGesture(perform: action)
+    }
+
+    @ViewBuilder
```
waliid:
```suggestion
```

Sources/CoreBusiness/Types/URNAssetLoader.swift
waliid, line 60:
```suggestion
//
//  Copyright (c) SRG SSR. All rights reserved.
//
//  License information is available from the LICENSE file.
//

import Combine
import PillarboxPlayer

enum URNAssetLoader: AssetLoader {
    struct Input {
        let urn: String
        let server: Server
        let configuration: PlaybackConfiguration
    }

    static func metadataPublisher(for input: Input) -> AnyPublisher<MediaMetadata, any Error> {
        let dataProvider = DataProvider(server: input.server)
        return dataProvider.mediaCompositionPublisher(forUrn: input.urn)
            .tryMap { response in
                try MediaMetadata(mediaCompositionResponse: response, dataProvider: dataProvider)
            }
            .eraseToAnyPublisher()
    }

    static func asset(from input: Input, metadata: MediaMetadata) -> Asset {
        if let blockingReason = metadata.blockingReason {
            return .unavailable(with: BlockingError(reason: blockingReason))
        }
        guard let resource = metadata.resource else {
            return .unavailable(with: SourceError())
        }
        let configuration = assetConfiguration(for: resource, configuration: input.configuration)
        if let certificateUrl = resource.drms.first(where: { $0.type == .fairPlay })?.certificateUrl {
            return .encrypted(url: resource.url, certificateUrl: certificateUrl, configuration: configuration)
        }
        else {
            switch resource.tokenType {
            case .akamai:
                return .tokenProtected(url: resource.url, configuration: configuration)
            default:
                return .simple(url: resource.url, configuration: configuration)
            }
        }
    }

    static func playerMetadata(from input: Input, metadata: MediaMetadata?) -> PlayerMetadata {
        metadata?.playerMetadata() ?? .empty
    }

    private static func assetConfiguration(
        for resource: MediaComposition.Resource,
        configuration: PlaybackConfiguration
    ) -> PlaybackConfiguration {
        // Limit buffering and force the player to return to the live edge when re-buffering. This ensures
        // livestreams cannot be paused and resumed in the past, as requested by business people.
        guard resource.streamType == .live else { return configuration }
        return .init(automaticallyPreservesTimeOffsetFromLive: true, preferredForwardBufferDuration: 1)
    }
}
```

Demo/Sources/Downloads/DemoAssetDownloadStore~ios.swift
defagos: Done on all files which were inconsistently ordered.

Demo/Sources/Downloads/DemoAssetDownloadStore~ios.swift
```diff
@@ -0,0 +1,127 @@
...
+    private struct FileEntry: Codable {
+        let id: String
+        let title: String
+        let url: URL
+        let metadata: String?
+        let bookmarkData: Data?
+        let progress: Double
+        let errorDescription: String?
+
+private init(id: String, title: String, url: URL, metadata: String?, bookmarkData: Data?, progress: Double, errorDescription: String?) {
```
waliid:
```suggestion
        private init(id: String, title: String, url: URL, metadata: String?, bookmarkData: Data?, progress: Double, errorDescription: String?) {
```

Sources/Player/Downloads/DownloadSession.swift
```diff
@@ -0,0 +1,44 @@
...
+@available(tvOS, unavailable)
+extension DownloadSession {
+    private static func downloadSessionTaskPublisher(for task: URLSessionTask) -> AnyPublisher<DownloadSessionTaskProperties, Never> {
+        Publishers.CombineLatest3(
+            Just(task),
+            task.publisher(for: \.state),
+            task.progress.publisher(for: \.fractionCompleted)
+                .map { $0.clamped(to: 0...1) }
+        )
+        .map { .init(task: $0, state: $1, progress: $2) }
```
waliid:
```suggestion
        .map(DownloadSessionTaskProperties.init)
```

Sources/Player/Downloads/DownloadSession.swift
```diff
@@ -0,0 +1,44 @@
...
+
+    func downloadSourcePublisher(
+        id: String,
+        asset: Asset,
+        title: String?,
+        createTaskIfNeeded: Bool,
+        progressEstimate: Double
+    ) -> AnyPublisher<DownloadSource, Never> {
+        sessionTaskPublisher(id: id, asset: asset, title: title, createIfNeeded: createTaskIfNeeded)
+            .map { Self.downloadSessionTaskPublisher(for: $0) }
```
waliid:
```suggestion
            .map(Self.downloadSessionTaskPublisher)
```

Sources/Player/Downloads/DownloadSession.swift
```diff
@@ -0,0 +1,44 @@
...
+            .switchToLatest()
+            .map { .task($0) }
```
waliid:
```suggestion
           .map(DownloadSource.task)
```

Sources/Player/Downloads/DownloadProperties.swift
```diff
@@ -0,0 +1,87 @@
...
+        do {
+            self.fileUrl = try URL(resolvingBookmarkData: record.bookmarkData)
+            self.source = .estimate(record.progress)
+            self.error = record.error
+        } catch {
+            self.fileUrl = nil
+            self.source = .estimate(0)
+            self.error = error
+        }
+    }
```
waliid, line 80:
```suggestion
    init<Input>(from record: DownloadRecord<Input, Metadata>) {
        do {
            self.init(
                metadata: record.metadata,
                source: .estimate(record.progress),
                fileUrl: try URL(resolvingBookmarkData: record.bookmarkData),
                error: record.error
            )
        } catch {
            self.init(metadata: record.metadata, source: .estimate(0), fileUrl: nil, error: error)
        }
    }
```

Sources/Player/Downloads/URLDownloadSession.swift
```diff
@@ -0,0 +1,76 @@
...
+                }
+            }
+            .eraseToAnyPublisher()
+    }
+
+    private func createTask(id: String, asset: Asset, title: String?) -> URLSessionTask {
+        let configuration = AVAssetDownloadConfiguration(
+            asset: asset.urlAsset(),
+            title: title ?? id
+        )
```
waliid:
```suggestion
        let configuration = AVAssetDownloadConfiguration(asset: asset.urlAsset(), title: title ?? id)
```

Sources/Player/ResourceLoading/Resource.swift
```diff
@@ -33,29 +37,31 @@ enum Resource {
...
+            return asset(for: url, with: configuration)
         case let .custom(url: url, delegate: delegate):
-            return ResourceLoadedPlayerItem(
-                asset: asset(for: url, with: configuration),
-                resourceLoaderDelegate: delegate,
-                automaticallyLoadedAssetKeys: automaticallyLoadedAssetKeys
-            )
+            let asset = asset(for: url, with: configuration)
+            asset.resourceLoader.setDelegate(delegate, queue: kResourceLoaderQueue)
+            objc_setAssociatedObject(asset, &kResourceLoaderDelegate, delegate, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
```
waliid:
```suggestion
```

Seen together, the `delegate` is already retained by the `AssetContent` which is itself retained by the `PlayerItem`.

Tests/CoreTests/FailPublisherTests.swift
```diff
@@ -0,0 +1,51 @@
...
+        let publisher = subject
+            .fail(onOutputFrom: signal, with: CancellationError())
+        let values = collectOutput(from: publisher, during: .milliseconds(100)) {
+            subject.send("A")
+            subject.send("B")
+            signal.send(())
+            subject.send("C")
+        }
+        expect(values).to(equalDiff(["A", "B"]))
+    }
```
waliid, line 50:
```suggestion
    }
    
    func testCompletionWithoutFailure() {
        let signal = PassthroughSubject<Void, Never>()
        let subject = PassthroughSubject<String, CancellationError>()
        let publisher = subject
            .fail(onOutputFrom: signal, with: CancellationError())

        expectSuccess(from: publisher) {
            subject.send(completion: .finished)
        }
    }
```

Sources/Core/Extensions/Publisher.swift
```diff
@@ -153,6 +153,22 @@ public extension Publisher {
...
+    /// - Returns: A publisher emitting values until the signal publisher emits a value, which leads to the publisher
+    ///   completing with an error.
+    func fail<S>(onOutputFrom signal: S, with error: Failure) -> AnyPublisher<Output, Failure> where S: Publisher, S.Failure == Never {
+        merge(
+            with: signal.first()
+                .map { _ in Fail<Output, Failure>(error: error) }
+                .switchToLatest()
+        )
+        .eraseToAnyPublisher()
+    }
```
waliid, line 175:
```suggestion
    func fail<S>(onOutputFrom signal: S, with error: Failure) -> AnyPublisher<Output, Failure> where S: Publisher, S.Failure == Never {
        let completionTrigger = PassthroughSubject<Void, Never>()
        return handleEvents(receiveCompletion: { _ in completionTrigger.send(()) })
            .merge(
                with: signal.first()
                    .prefix(untilOutputFrom: completionTrigger)
                    .map { _ in Fail<Output, Failure>(error: error) }
                    .switchToLatest()
            )
            .eraseToAnyPublisher()
    }
```

waliid (approved):

waliid (approved):

Issue #1499, opened 2026-05-31
Offline simple asset support


defagos:
**As a developer integrating Pillarbox** with basic needs I would like to be able to quickly download simple assets without having to implement full-fledged support for custom metadata persistence.

## Acceptance criteria

- An API is provided to download simple URLs with simple metadata (e.g. `PlayerMetadata`), with which adding downloads to an app is a matter of minutes.

## Tasks

- [x] Move `URLAssetDownloadStore` to PillarboxPlayer.
- [x] Provide public `URLDownloader` based on internal `URLAssetDownloadStore`.
- [x] Support custom data if possible (probably needs to be `Codable`).
- [x] Check if/how void custom data deserves convenience APIs like those of `PlayerItem` (using `PlayerMetadata` directly) and if, in general, such APIs make sense.
- [x] Support custom and encrypted assets if possible.
- [x] Handle/persist playback configuration.
- [x] Remove old unused files, if any.
- [x] Update implementation to use SwiftData.
- [x] Update the demo to work with the new `URLAssetDownloadStore`.
- [x] Make sure the download feature is not visible on iOS 16.
- [x] ~Custom store implementations are still possible (asset loader needs to be exposed?).~
- [x] Keep only relevant parameters (URN, URL) in download retrieval methods on `URNDownloader` and `URLDownloader`.
- [x] Avoid issues with URN/URL SwiftData stores potentially sharing the same name, or with model class names potentially clashing.

defagos:
The (playback) configuration does not require special handling or persistence. Since it is created from metadata in the general case, it is only natural that we do the same from custom data in the case of URL-based asset resource loading.

defagos:
I checked whether exposing `URLAssetLoader` publicly to allow custom store implementations was useful. It isn't since the provider mechanism we introduced for customization makes implementations more difficult than they should be.

Instead, custom URL-based content downloading should likely be implemented from scratch (the asset loader itself is trivial and can be tailored to the exact metadata required).

PR #1512, opened 2026-06-21
Add offline support for Core Business URN-based content 

### LINKED ISSUE #1452: Offline Core Business metadata management
## Description

**As a developer integrating Pillarbox** I would like to be able to download URN-based content provided by Core Business.

## Hints

See #1115 for preliminary investigations.

## Acceptance criteria

- URNs can be downloaded.
- Metadata is displayed offline.
- Blocked time ranges cannot be played offline.
- Chapters are displayed offline.
- Tracking works when playing offline.
- All content in the demo offers a download button (some content might not be playable offline yet, e.g. live streams or DRM-protected content, though) on the content cells.
- A list displays all content available offline.
- Error for non-playable content (e.g. live streams) are correctly reported.

Not in scope:

- DRM/token-protected content is not supported yet (done in #1431).
- Images are not stored locally yet (including the artwork associated with the content in the Settings app).

## Tasks

- [x] Implement asset download store for media composition metadata using Swift Data.
    - [x] Check Swift Data SQLite store to ensure that no entries are correctly written (no duplicates, correct cleanup on deletion) on all supported versions of iOS 17+.
    - [x] Encapsulate instantiation so that it does not require any setup from the app (especially not setting required models or migration plans). Validate by having two URN stores in the same app.
- [x] Fix crash with EpisodeInformation (e.g. extract from 19h30 downloaded, crashes when played).
- [x] Add subtitle on download cell.
- [x] Display download list in the demo.
- [x] Add download button to content cells.
- [ ] Provide download helpers if meaningful (e.g. specialized `addDownload` API), also for direct URL-based downloads if possible.
- [ ] Check public API surface.
- [x] ~Fix/improve following scenario if possible (+UT):~
    - Download file when running as iPad app on macOS.
    - Delete file but keep in Trash.
    - Relaunch app. Bookmark is still valid but file is not playable. Bookmark is nilled.
    - Empty Trash and relaunch app. The download should be in a _Missing file_ error state (recoverable manually) but isn't.

defagos:
## Description

This PR implements offline playback for SRG SSR content delivered by our Core Business framework.

### Remark

Content protected with DRMs, tokens or AES encryption will be addressed separately.

## Design considerations

- We kept the `AssetLoader` protocol as is, i.e. it is still responsible of delivering rich metadata, from which assets can be created and with which trackers can be supplied the information they require (if available). We only removed a default implementation when metadata is `PlayerMetadata` (this case is an exception, the default implementation was not justified).
- We focused on updating `AssetDownloadStore` so that rich metadata from the asset loader can be sliced as required for convenient persistence. This avoids the need to store all data from the asset loader, which might be cumbersome depending on what is delivered.
- We linked `AssetDownloadStore` with an asset loader through an associated type. This makes sense, as stores will always have a strong coupling with the asset loader they get data from (most notably the input will always need to be identical).
- We revisited player item creation helpers. To provide for maximum flexibility in how code can be reused, custom data is supported for simple, custom and encrypted items. Default values have been adjusted accordingly. For example, supplying custom data is only meaningful for trackers at the `PlayerItem` creation level. For this reason having a default tracker list does not make sense. On the other hand, trackers are less likely to be useful when metadata is `PlayerMetadata`, since relevant data might be missing. For this reason a default empty tracker list is provided for corresponding creation helpers.

## Changes made

- Add `URNDownloader` to manage SRG SSR downloads.
- Make `PlayerMetadata` and associated types conform to `Codable` for easy persistence. `PlayerMetadata` is not directly storable as a `Codable` object in Swift Data because of Swift Data limitations and `EpisodeInformation` requires a trick to be properly stored. We'll investigate if this can be improved in #1513.
- Update `PlayerItem` creation helpers to support custom data.
- Support URL-based 360° monoscopic content. Time ranges and chapters are not implemented yet, as extensive support will be added in #1499.
- Update demo to support downloads for URLs and URNs via swipe actions.
- (Non-related) Add `LogForEeachSlowPath` to available demo environment variables for slow UI update detection.
- Use Homebrew to install pkgx, circumventing [recent issue](https://github.com/pkgxdev/pantry/issues/13490) with installation via curl.

## Checklist


Sources/Player/Types/EpisodeInformation.swift
```diff
@@ -5,10 +5,16 @@
 //
 
 /// Information about an episode.
-public enum EpisodeInformation: Equatable {
-    /// Season and episode number.
-    case long(season: Int, episode: Int)
+public struct EpisodeInformation: Codable, Equatable {
+    /// Season number.
```
waliid:
```suggestion
    /// Episode number.
```

Sources/Player/Types/EpisodeInformation.swift
```diff
@@ -5,10 +5,16 @@
...
+    let episode: Int
 
-    /// Episode number only.
-    case short(episode: Int)
+    /// Episode number.
```
waliid, line 9:
```suggestion
    /// Season number.
```

waliid (approved):

PR #1526, opened 2026-06-26
Add offline image persistence for downloads

### LINKED ISSUE #1507: Offline image management
**As an app user** I want the content I have downloaded to display images, even when I am offline:

- In the Control Center.
- When chapters are displayed.
- When an artwork is displayed in the app (e.g. player background).
- In the system storage settings.

## Acceptance criteria

- The API provides a way to save/load images from local storage.
- Existing asset loaders have been updated to support images.
- The demo displays images for downloaded content (see use cases above).

## Tasks

- [x] Lazily load images on the custom player view.
- [x] Make image source publisher resilient so that it can be re-attempted if needed.
- [x] Improve current implementation so that:
    - [x] Main artwork retrieval is resilient (i.e. if it failed it can still be loaded when possible).
    - [x] Chapter images that could not be loaded can be loaded again.
- [x] Update metadata and download properties publisher so that:
    - [x] The artwork (if any) is retrieved before the task is created (assign it to the task). In case of image retrieval failure continue with no image.
    - [x] Load chapter images.
    - [x] Ensure metadata retrieval is consistent both when initially loading metadata remotely as well as when metadata is available locally.
- [x] Update `ImageSource.image(...)` to associate `Data` instead of `UIImage`?
- [x] Use JPEG data for images (check result in Control Center and on Apple TV).
- [x] Update URN store to persist image data (possibly externally).
- [x] Test behavior with poor/unreliable network conditions.
- [x] Check storage (correct cleanup on download removal, size, etc.).
- [x] Run the import/dead code cleaner :)

waliid:
## Description

This PR adds offline image persistence for downloaded content.

## Design considerations

- Download asset construction has been moved to the store implementation. This is more flexible than having the asset loader force a decision. Several download options might namely be available and a store implementation might want to pick the optimal one.
- To load images we have applied a similar strategy to the one applied until now to load images into the player, though without lazy loading mechanism.
- Image loading publishers have been made more resilient so that images which failed to load (either when playing the content online or when downloading it) can be loaded again later.
- Lazy loading has been separated and is not a byproduct of reading an image from an image source anymore. This allows us to more finely control fetch attempts, thus allowing graceful image loading retrieval to be performed on view appearance (e.g. as a result of a scrolling operation).
- Image sources now have attached `Data` instead of a `UIImage`. Not only does this remove a dubious dependency to UIKit, it also lets implementations choose whether they prefer PNG or JPEG when storing images (including how much compression they desire).

## Changes made

- A `lazyImageSourcePublisher` has introduced to allow retrieving image on demand.
- In the demo app the `HStack` has been replaced by a `LazyHStack` for the chapters list to allow redrawing.
- The download has been updated so that main artwork is retrieved before task creation.
- The download has beeen updated so that chapters images are fetched after task creation.
- The asset publisher must be implemented when implementing `AssetDownloadStore`.
- `ImageSource` has been updated so that a raw `Data` is associated instead of `UIImage`.
- The URN store has updated to allow image url and image data to be persisted.
- Imports have been cleaned up.

## Checklist


Demo/Sources/Downloads/DownloadAction~ios.swift
```diff
@@ -0,0 +1,24 @@
...
+        if downloader.canDownload(media: media) {
+            Button {
+                downloader.addDownload(media: media)
+            } label: {
+                Image(systemName: "arrow.down.circle")
+            }
+            .tint(.green)
+        }
+    }
+}
```
defagos, line 26: The demo does not compile for the release configuration. To fix:

```suggestion
//

#if DEBUG

import SwiftUI

struct DownloadAction: View {
    let media: Media

    @EnvironmentObject private var downloader: DemoDownloader

    var body: some View {
        if downloader.canDownload(media: media) {
            Button {
                downloader.addDownload(media: media)
            } label: {
                Image(systemName: "arrow.down.circle")
            }
            .tint(.green)
        }
    }
}

#endif
```

Sources/Player/Downloads/Download.swift
```diff
@@ -157,25 +156,28 @@ private extension Download {
         if !properties.shouldCreateTask, let metadata = properties.source.metadata {
             return session.sessionTaskPublisher(id: id)
                 .setFailureType(to: Error.self)
-                .map { session.downloadSessionTaskPropertiesPublisher(for: $0) }
+                .map { task in
+                    Publishers.CombineLatest(
+                        session.downloadSourceTaskPublisher(for: task, using: properties),
```
defagos:
```suggestion
                        session.downloadSourceTaskPublisher(for: task, properties: properties),
```

Sources/Player/Downloads/DownloadSession.swift
```diff
@@ -29,6 +29,16 @@ extension DownloadSession {
         .map(DownloadSessionTaskProperties.init)
         .eraseToAnyPublisher()
     }
+
+    func downloadSourceTaskPublisher<CustomData>(
+        for task: URLSessionTask?,
+        using properties: DownloadProperties<CustomData>
```
defagos:
```suggestion
        properties: DownloadProperties<CustomData>
```

Sources/Player/Downloads/DownloadSession.swift
```diff
@@ -29,6 +29,16 @@ extension DownloadSession {
...
+    ) -> AnyPublisher<DownloadSourceKind, Never> {
+        guard let task else { return Just(.estimate(properties.progress)).eraseToAnyPublisher() }
```
defagos, line 37: I am asking myself if we could end up in a state for which:

- There is a file partially downloaded (progress < 1).
- There is no task anymore.

Not a problem for the moment but we should keep an eye, either on downloads potentially stuck at some progress < 1, or at downloads for which a dangling file could remain.

defagos (approved):

defagos (approved):

PR #1535, opened 2026-06-30
Preserve metadata when restarting a download

### LINKED ISSUE #1529: Preserve metadata when restarting a download
**As a user** I find it disturbing that restarting a download temporarily loses the associated title (which was already known).

## Acceptance criteria

- When restarting a download in the demo, and provided its metadata was available, the metadata is not temporarily reset anymore.

## Tasks

- [x] Update download pipeline to restart a download with existing metadata, if any.
- [x] Add test for restart right after download has been added. Must work.
- [x] Make `Download.remove()` internal.

waliid:
## Description

This PR allows us to keep the metadata previously retrieved when restarting a download.

## Changes made

- A `resetRecord` closure  has been added to update a download record without removing its previously retrieved metadata.
- The `downloadPropertiesPublisher` has been updated to prepend the previous metadata and also using it when catching an error.

## Checklist

### DIFF
--- a/Sources/Player/Downloads/Download.swift
+++ b/Sources/Player/Downloads/Download.swift
@@ -203,14 +203,7 @@ private extension Download {
                 }
                 .switchToLatest()
                 .fail(onOutputFrom: trigger.signal(activatedBy: TriggerId.cancel), with: URLError(.cancelled))
-                .catch { error in
-                    Just(DownloadProperties(
-                        progress: .estimate(0),
-                        assetMetadata: storedProperties.assetMetadata,
-                        fileUrl: storedProperties.fileUrl,
-                        error: error
-                    ))
-                }
+                .catch { Just(store.downloadProperties(forId: id).withError($0)) }
                 .prepend(storedProperties)
         }
     }
--- a/Sources/Player/Downloads/Download.swift
+++ b/Sources/Player/Downloads/Download.swift
@@ -24,7 +24,7 @@ public final class Download: ObservableObject {
     private let trigger = Trigger()
     private let session: any DownloadSession
 
-    private let resetRecord: (Bool) -> Void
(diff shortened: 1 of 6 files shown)


Sources/Player/Downloads/Download.swift
```diff
@@ -258,7 +264,10 @@ extension Download {
             )
             .map { DownloadProperties(source: $0, fileUrl: $1, error: $2) }
             .fail(onOutputFrom: trigger.signal(activatedBy: TriggerId.cancel), with: URLError(.cancelled))
-            .catch { Just(DownloadProperties(source: .init(kind: .estimate(0), metadata: nil), fileUrl: nil, error: $0)) }
+            .catch { error in
+                let properties = store.downloadProperties(forId: id)
+                return Just(DownloadProperties(source: properties.source, fileUrl: nil, error: error))
+            }
```
defagos: The fact that we lose previous property updates in case of an error, forcing retrieval from current values, highlights a potential flaw with our current implementation. The pipeline should namely not require state to be recovered externally but should be able to deliver all properties on its own.

Instead, I would suggest we try delivering a property **update** stream (probably 4 cases: task, metadata, location, error), with a scan consolidating updates into properties before delivering the updated properties for persistence. This way previous values would never be lost and we would likely be able to eliminate the store access already made when restarting a download.

Maybe this could also make our internal types simpler. I am not sure that `DownloadSource`, in particular, would then be required anymore.

defagos (approved):

Sources/Player/Downloads/Download.swift
defagos: Not a flaw in our design. The record is altered through the publisher and access to the store is only made sparingly (in the event of an error).

defagos (approved):

Issue #1379, opened 2025-11-25
The tvOS player UI leaks memory on tvOS 26


defagos:
### Is there an existing issue for this?

- [x] I have searched existing issues and found no similar bug report.

### Description of the problem

Starting with tvOS 26, the standard `SystemVideoView` leaks memory for content containing chapters.

### Relevant stack trace or log output

```shell
None
```

### Reproducibility

Always

### Steps to reproduce

1. Open a 19h30 in the demo, close it.
2. Repeat and observe memory growth (or repeat until the system kills the demo app).

### Library version

13.0.1

### Operating system

tvOS 26.1

### Code sample

None

defagos:
We got feedback from Apple that this issue might have been fixed in tvOS 27 beta 1. Sadly this is not the case. We answered their feedback and we hope this will truly be the case in upcoming beta versions.

defagos:
The issue has been fixed in tvOS 26.3 beta 3. The radar has been updated and closed.

PR #1538, opened 2026-07-04
Avoid chapter artwork leak on tvOS 26

defagos:
## Description

This PR offers a workaround for the tvOS chapter artwork leak described in #1379. It implements a completely custom chapters panel that mimics the system one on tvOS 26, with a few minor differences:

- Selecting a chapter does not close the info panel automatically.
- There is no marquee effect on chapter titles. This effect is tricky to implement and probably too much work for what is a temporary workaround anyway. A nice [library](https://github.com/GRimAce11/MarqueeKit) exist but this would add a dependency for a very specific need.

result

Note that the workaround has been applied on tvOS 26 only since tvOS 27 will likely fix the leak. We were namely informed of a change in beta 1 and can confirm that the leak has been reduce, though not entirely eliminated yet.

Also note that the tvOS panel APIs have never been marked as tvOS-specific. For this reason the new chapter UI types have not been marked as unavailable on iOS.

### Remark

Accessibility is not perfect, as tab titles are appended for some reason to be determined. I have created a dedicated task since this likely affects all custom info panels.

## Changes made

- Disable navigation markers on tvOS 26.
- Implement chapters tab replacement for tvOS 26, based on the design for this version (the design was updated in tvOS 27). The blurry image extension was obtained using the new background extension effect introduced in tvOS 26.
- Improve `Tab` initializers for localization. The `title` parameter was removed to match usual SwiftUI tab creation.
- Remove associated known issue.
- Translate all required strings. Those have been borrowed from tvOS directly (except for the Romansh one) and are updated on Crowdin.

## Checklist

### DIFF
--- a/Demo/Sources/Views/MediaCardView~tvos.swift
+++ b/Demo/Sources/Views/MediaCardView~tvos.swift
@@ -85,11 +85,7 @@ struct MediaCardView: View {
                 }
             }
             .overlay {
-                LinearGradient(
-                    colors: [.black, .clear],
-                    startPoint: .bottom,
-                    endPoint: .top
-                )
+                LinearGradient(colors: [.black, .clear], startPoint: .bottom, endPoint: .top)
             }
             .frame(width: size.width, height: size.height, alignment: .center)
 
--- a/Demo/Sources/Views/MediaCardView~tvos.swift
+++ b/Demo/Sources/Views/MediaCardView~tvos.swift
@@ -85,7 +85,11 @@ struct MediaCardView: View {
                 }
             }
             .overlay {
-                LinearGradient(gradient: Gradient(colors: [Color.clear, Color.black]), startPoint: .top, endPoint: .bottom)
+                LinearGradient(
+                    colors: [.black, .clear],
+                    startPoint: .bottom,
+                    endPoint: .top
+                )
             }
             .frame(width: size.width, height: size.height, alignment: .center)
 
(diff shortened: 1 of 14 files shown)


Demo/Sources/Views/MediaCardView~tvos.swift
```diff
@@ -85,7 +85,11 @@ struct MediaCardView: View {
                 }
             }
             .overlay {
-                LinearGradient(gradient: Gradient(colors: [Color.clear, Color.black]), startPoint: .top, endPoint: .bottom)
+                LinearGradient(
+                    colors: [.black, .clear],
+                    startPoint: .bottom,
+                    endPoint: .top
+                )
```
waliid:
```suggestion
                LinearGradient(colors: [.black, .clear], startPoint: .bottom, endPoint: .top)
```

Sources/Player/UserInterface/ChapterCell.swift
```diff
@@ -0,0 +1,105 @@
...
+                .aspectRatio(Self.aspectRatio, contentMode: .fit)
+                .backgroundExtension(spacing: Self.heightExtension)
+        }
+        .frame(maxWidth: .infinity, maxHeight: .infinity)
+        .overlay(
+            LinearGradient(
+                colors: [.black, .clear],
+                startPoint: .bottom,
+                endPoint: .center
+            )
```
waliid:
```suggestion
            LinearGradient(colors: [.black, .clear], startPoint: .bottom, endPoint: .center)
```

Demo/Sources/Views/MediaCardView~tvos.swift
defagos: Thanks. I turned all `LinearGradient` into one-liners.

waliid (approved):

PR #1566, opened 2026-08-21
Support quality and media selection choice for downloaded content

### LINKED ISSUE #1434: Offline rendition management
**As a user** I would like to be able to control which subtitles, audio tracks and qualities are downloaded so that I can optimize download size and device free space.

### Remarks

- The system player UI seems to filter out locally unavailable renditions, our `Player` APIs currently probably don't and will provide all renditions from offline persisted playlists, even if these renditions have not been downloaded.
- We should ensure the player never attempts loading unavailable renditions, otherwise playback will start with a "No network" error. To test:
    - Download a content with only selected subtitles downloaded.
    - Pick subtitles that are not available. If online they will be displayed and content will play just fine.
    - Close the player and disconnect the network. When opening the player previous subtitles will be restored. If not available and incorrectly not filtered, playback fails with a "No network" error.

## Hints

- [WWDC 2021 session](https://developer.apple.com/videos/play/wwdc2021/10143/) (configuration API)
- [WWDC 2020 session](https://developer.apple.com/videos/play/wwdc2020/10655) (offline playback in general but older API)

## Acceptance criteria

- An API to select renditions is available.
- Automatic selection is correctly taken into account.
- AD and SDH preferences are correctly taken into account.
- Rendition selection is possible based on media selections and variant qualifiers.
- Renditions can be managed if possible (e.g. later download another rendition, delete a rendition).
- Available renditions for a downloaded file can be listed and displayed in the demo (if possible).
- The demo provides a way to customize rendition selection (e.g. global setting or at the time a download is started) or management if possible (e.g. from a dedicated contextual menu).

## Tasks

- [x] Add download configuration parameter.
- [x] Add download configuration to the persistence layer.
- [x] Provide a way to set a maximum bit rate.
- [x] Make sure that only options available offline are returned by `Player` media selection APIs (via `AVAssetCache`) when playing a local file. Do not return other options.
- [x] Support loose language codes (first check for a strict match, then use code without the country). This is how media selection criteria work. Check if we should loosen language matching with several priorities, possibly optionally (e.g. `fr-FR` matches `fr-FR` exactly, fr less and `fr-CA` even less). Check how the system behaves first. Maybe having `fr` select all `fr*` is also a viable option.
- [x] Provide a way to choose primary and auxiliary (if needed) audio and subtitles selections.
- [x] Deal with forced subtitles:
    - [x] Download unforced subtitles when setting legible languages.
    - [x] Download forced subtitles automatically if possible (for all audio languages requested).
- [x] Download SDH subtitles preferably if the corresponding a11y option is enabled (no work required for automatic mode, only if a specific selection is requested).
- [x] Download AD preferably if the corresponding a11y option is enabled (no work required for automatic mode, only if a specific selection is requested).
- [x] Add option to download all variants.
- [x] Add support for resolution setting as well.
- [x] Implement demo settings:
    - [x] Quality.
    - [x] Audio.
    - [x] Subtitles.
- [x] ~Add a way to get downloaded subtitle and audio options. Should probably not be automatic (since requires the asset to be checked). E.g. a `DownloadInspector` observable objec that will only load tracks lazily for download cells actually displayed.~
- [x] Check with URL/URN content.
- [x] Check behavior on iOS 16+.
- [x] Ensure a download is always playable, even if starting with a language that could be supported but was not downloaded. Might not be possible, in which case report media selection/offline bug to Apple and add to known issues.
- [x] Report system player UI issue to Apple: Only variants available offline must be displayed.
- [x] Add reported issues to known issues.

defagos:
## Description

This PR adds an optional configuration that can be supplied when starting (or restarting) a download. This configuration can be used to:

- Control the maximum quality of the downloaded stream, either via bitrate or resolution upper bounds.
- Declare which audio and subtitles should be downloaded if possible. Options are available to select languages explicitly by language code, download all available variants or fallback on default automatic behavior.

result

## Design and implementation considerations

- The configuration is provided when starting a download. By default a download is made with the default configuration (no quality limit, default media selection based on device language and a11y settings). When restarting a download either the original configuration can be reapplied or a new one supplied. This approach allows apps to decide whether they want to associate settings with a download permanently or apply most recent settings (possibly globally managed at the app level or requested from the user at the time the download is started or restarted).
- The configuration has been added to a download record and must therefore be persisted to storage. This is required for downloads to be able to reuse their initial configuration. It also fits within the implementation more naturally (retrieval through initial stored properties, from which the configuration can be read afterwards in the Combine download pipeline we have).
- The new `DownloadMediaSelectionPreference` is very similar to our existing `MediaSelectionPreference`. A separate type was introduced, though, as options differ a bit.
- Configuration APIs have been made more convenient as in a corresponding [Castor PR](https://github.com/SRGSSR/castor/pull/298).
- Forced subtitles are automatically retrieved for each downloaded audio. Note that having both the audible and forced legible options set on the same `AVMediaSelection` per language seems required for proper download behavior (some downloads namely would sometimes get stuck if two separate media selections were used).
- AD and SDH subtitles are automatically retrieved based on a11y settings, aligned with the default automatic behavior.
- When playing an offline asset, only audio tracks and subtitles available offline are returned. This is not aligned with the system player user experience but solves related issues (reported as FB24446455).

## Known issues

- Offline playback fails due to media selection picking variants declared in the master playlist but not available offline (FB24445920). This issue is rather annoying but no safe mitigation exists IMHO. Messing with automatic selection or attempting to force an option is namely the shortest path to disaster, as behaviors are quite subtle and affect all kinds of context (simple playback, playlist, AirPlay). Until Apple's engineers fix this issue apps can decide to download `.all` options (safest approach) or use automatic mode (quite safe in general). Providing an explicit language works as well but I would rather recommend this option for apps delivering content consistently with the same audio and subtitle options.
- The system video view might present options unavailable for offline playback (FB24446455).

## Changes made

- Implemented design described above.
- Updated standard URN downloader to support configurations as well.
- Updated configuration objects to be more ergonomic. The new `DownloadConfiguration` does not offer a media selection per characteristic in its constructor, though, as this is not really ergonomic.
- Added download settings on iOS (quality, audio, subtitles). Audio and subtitles provide an option for downloading the CH language list (a free language choice was considered but not really practical). Starting or restarting a download always retrieves the most recent settings from user defaults.
- Updated demo settings to remove those not relevant to tvOS on this platform.
- No tests were sadly written since downloads require `AVURLDownloadSession`. Most of the code was written in a way that blends with our existing online track management code, though.

## How to test

- Use the new quality setting to set the resolution to low. This makes downloads end very fast.
- Use _The Morning Show_ for most tests. It offers the largest audio and subtitle selection.
- Play the languages and a11y settings and observe changes after (re)starting a download.
- For forced subtitles [a proxy](https://github.com/SRGSSR/pillarbox-apple/issues/1434#issuecomment-5338482901) is needed to make Apple's Bip Bop work as expected.

## Checklist


Sources/Player/Downloads/DownloadConfiguration.swift
```diff
@@ -0,0 +1,84 @@
...
+                        AVAssetVariantQualifier.predicate(forPresentationWidth: preferredMaximumResolutionWidth, operatorType: .lessThanOrEqualTo)
+                    ]
+                )
+            )
+        ]
+        if let selection {
+            configuration.primaryContentConfiguration.mediaSelections = mediaSelections(from: selection, using: provider)
+        }
+        configuration.auxiliaryContentConfigurations = []
+    }
```
waliid, line 76:
```suggestion
    func apply(selection: AVMediaSelection?, to configuration: AVAssetDownloadConfiguration, using provider: MediaSelectionProvider) {
        let resolutionPredicate = NSCompoundPredicate(andPredicateWithSubpredicates: [
            AVAssetVariantQualifier.predicate(forPresentationWidth: preferredMaximumResolutionWidth, operatorType: .lessThanOrEqualTo),
            AVAssetVariantQualifier.predicate(forPresentationHeight: preferredMaximumResolutionHeight, operatorType: .lessThanOrEqualTo)
        ])
        let peakBitRatePredicate = NSPredicate(format: "peakBitRate <= \(preferredPeakBitRate)")
        configuration.primaryContentConfiguration.variantQualifiers = [
            AVAssetVariantQualifier(predicate: NSCompoundPredicate(orPredicateWithSubpredicates: [resolutionPredicate, peakBitRatePredicate]))
        ]
        if let selection {
            configuration.primaryContentConfiguration.mediaSelections = mediaSelections(from: selection, using: provider)
        }
        configuration.auxiliaryContentConfigurations = []
    }
```

Sources/Player/Downloads/Download.swift
```diff
@@ -99,9 +117,17 @@ public extension Download {
     }
 
     func restart() {
+        _restart(configuration: nil)
```
waliid:
```suggestion
        _restart(configuration: configuration)
```

Should we keep the initial configuration?

Sources/Player/Downloads/DownloadConfiguration.swift
defagos: The `AVPlayer.preferredMaximumResolution` and `AVPlayer. preferredPeakBitRate` work in an `OR` fashion as well. You proposal clearly makes sense, thanks.

waliid (approved):

PR #1567, opened 2026-08-22
Update for Xcode, iOS and tvOS 27

defagos:
## Description

This PR updates the project for optimal compatibility with the 27 releases.

## Hints

- `ContentBuilder` [in general](https://fatbobman.com/en/posts/contentbuilder-explained/).
- `ContentBuilder` [source-breaking changes](https://developer.apple.com/documentation/technotes/tn3211-resolving-swiftui-source-incompatibilities-for-state-and-contentbuilder).

## Changes made

- Updated Swift tools to version 6.4 to benefit from recent Swift language features.
- Modernized the code for Swift 6.4.
- Fixed implementation to account for periodic time observation change in 27 releases (see below).
- Fixed unreliable item cleanup on OS 27 releases, which was making the `CurrentItemTests.testSetCurrentItemToNil()` fail. In this case cleanup does not work once anymore, so we do it twice. 🤡
- Replaced `ViewBuilder` with `ContentBuilder`. Disambiguate types [where needed](https://developer.apple.com/documentation/technotes/tn3211-resolving-swiftui-source-incompatibilities-for-state-and-contentbuilder#Disambiguate-types-and-members-that-collide-across-modules).
- Fix option selection on tvOS 27. Since `.navigationLink` [does not work](https://github.com/SRGSSR/apple-bug-reports/blob/main/FB24831195/report.md) we wrap pickers into menus. This works better than using the `.menu` selection style, which only displays the selection value and not its label when collapsed.
- Updated CI tool chain for Xcode 27.

## Periodic time observation change

Several UTs were failing, in particular tests related to time-sensitive values at the very start of playback. After careful inspection all issues could be pinned to a single source, namely a change in how `AVPlayer.addPeriodicTimeObserver(forInterval:queue:)` behaves, both in the simulator and on device:

- Prior to OS 27 releases, an initial time would be emitted, even if the player was prepared in a paused state.
- Starting with OS 27 releases, no initial time is emitted in this case. It is only after the player starts playing at a non-zero rate that times start to be emitted.

On device, and with OS 27 final and 27.2 beta, it seems the issue on device depends on how fast the content is loaded  initially. The issue might not appear with a good connection but only if using the network link conditioner to simulate a poor network.

This new behavior messes up with several of our implementations which rely on an initial value being published (progress tracking, player item tracking, for example) and leads to the test failures mentioned above.

Since progress tracking is affected this can also have visible effects. Fir example, if we remove the call to `play()` in our `PlayerViewModel`, the custom player UI progress bar will not be visible on iOS 27 until playback starts, whereas this was the case on iOS 26 and below.

Here is a [sample project](https://github.com/user-attachments/files/31473971/PeriodicPublisher.zip) that shows this difference in behavior.

Simply run the project on iOS 26 (or below) and observe that an (invalid) time is logged to the console. Then do the same on iOS 27 and observe that no time is logged until the _Play_ button is tapped.

Note that no `VideoPlayer` was not used to avoid affecting time observation behavior. Indeed, if a `VideoPlayer` is used, no initial event will be emitted. This means that iOS 27 behavior is more consistent, i.e. no initial event is sent by a periodic publisher, whether a `VideoPlayer` is used or not.

### Conclusion

Whether this change is a regression or a fix is not easy to tell based on the documentation (there is nothing in the release notes either:

> The system invokes the block periodically at the interval specified, interpreted according to the timeline of the current item. It also invokes the block whenever time jumps or playback starts or stops.

If _starts_ means that playback moves forward, then this is actually a fix, since previously an event was emitted when playback was still paused.

The fact that the iOS 26 and 27 are now aligned, whether `VideoPlayer` is used or not, by inhibiting the first event, rather points at a behavioral fix. For this reason we should likely updated our implementation and avoid reporting this issue to Apple as a regression.

### Implementation update

The simplest possible approach is likely to prepend the player current time once to our periodic publisher. This requires a single test to be adjusted (initial invalid time emitted by an empty player) but fixes all other tests and behaviors.

This fix can also be applied consistently to older OS versions as well.

## Checklist

- [x] The behavior works with all receivers available in the demo.


defagos:
Apple introduced a regression with `Menu` and associated resource memory management. This has been reported as _iOS 27 regression: Objects whose properties are bound to items displayed in a `Menu` are not correctly deallocated_, filed under [FB24486991](https://github.com/SRGSSR/apple-bug-reports/tree/main/FB24486991).

A [message](https://developer.apple.com/forums/thread/843256) has been posted on the Apple Developer Forums as well.

Sources/Player/Player/Player+ItemNavigation.swift
```diff
@@ -133,7 +133,10 @@ extension Player {
             }
         }
         else {
+            // Cleaning once does not suffice anymore on OS 27 releases.
             queuePlayer.removeAllItems()
+            queuePlayer.removeAllItems()
+            print(queuePlayer.items())
```
waliid:
```suggestion
```

defagos:
The `CurrentItemTests.testSetCurrentItemToNil()` is a race condition. The test:

```swift
func testSetCurrentItemToNil() {
    let item = PlayerItem.simple(url: Stream.onDemand.url)
    let player = Player(item: item)
    expect(player.currentItem).to(equal(item))
    player.currentItem = nil
    expect(player.currentItem).to(beNil())
    expect(player.items).to(equalDiff([item]))
    expect(player.queuePlayer.items()).to(beEmpty())
}
```

namely fails on iOS 27 (succeeds on iOS 26) but the following succeeds on iOS 27:

```swift
func testSetCurrentItemToNil() {
    let item = PlayerItem.simple(url: Stream.onDemand.url)
    let player = Player(item: item)
    expect(player.currentItem).toAlways(equal(item), until: .seconds(1))
    player.currentItem = nil
    expect(player.currentItem).to(beNil())
    expect(player.items).to(equalDiff([item]))
    expect(player.queuePlayer.items()).to(beEmpty())
}
```

i.e. waiting a bit after initial item setup and before erasing all items suffices to make the test reliable again.

I tried to find other options than the duplicate removal to address this timing issue but couldn't find one for the moment.

waliid (approved):

PR #1568, opened 2026-08-22
Hide downloads behind preview trait

defagos:
# Description

This PR introduces a `DownloadPreview` trait with which downloads can be easily integrated as a feature preview. This official mechanism replaces the `@_spi(DownloaderPrivate)` attribute we were previously using.

## Hints

Some documentation about package traits:

- https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/packagetraits/
- https://www.massicotte.org/blog/package-traits-in-xcode/
- https://theswiftdev.com/all-about-swift-package-manager-traits/
- https://www.pointfree.co/blog/posts/216-trait-ifying-our-libraries-to-reduce-transitive-dependencies

## Changes made

- Introduced `DownloadPreview` package trait in manifest (name inspired from Swift Collection package traits).
- Removed `@_spi(DownloaderPrivate)`.
- Updated demo to use `DownloadPreview` trait instead of `@_spi(DownloaderPrivate)`. This required adding the local package explicitly as dependency (previously this dependency was implicit).
- Enabled downloads in all demo builds except release builds, using a dedicated preprocessor flag.

## Checklist

### DIFF
--- a/Demo/Info.plist
+++ b/Demo/Info.plist
@@ -2,8 +2,6 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
-	<key>ITSAppUsesNonExemptEncryption</key>
-	<false/>
 	<key>LSApplicationQueriesSchemes</key>
 	<array>
 		<string>github</string>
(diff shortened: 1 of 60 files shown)


defagos:
An Xcode bug seems to always clear traits enabled for **local** packages when opening a project. The project can be reverted to its prior state wth Xcode still open, though. Moreover there is no trait support for `xcodebuild`, thus download tests are omitted. 

For all these reasons this PR does not seem viable at the moment.

PR #1575, opened 2026-08-27
Fix download test stability

defagos:
## Description

This PR fixes download test stability when repeating the following tests several times:

- `DownloadTests.testRemoveWhileInitiallyRunning()`
- `DownloadTests.testRemoveWhileDownloadingFile()`

Often the `fileUrl` store in properties was not properly cleaned up after cancellation. We namely have two error management pipelines:

- One is for the download task itself, where locations and errors are received separately and not mutually exclusive (we can have an error **and** a location).
- The other one is for metadata retrieval and cancellation errors, for which location and error are mutually exclusive. In this pipeline we namely either never started the download task (i.e. we never had a chance to receive a `fileUrl`) or we cancelled the download somehow.

## Changes made

- Erased `fileUrl` stored in properties when updated with an error.
- Updated Couleur 3 DVR and vertical video examples.
- Updated Commanders Act SDK to version 5.5.1 (prior: 5.5.0).

## Checklist

### DIFF
--- a/Demo/Sources/Model/URNMedia.swift
+++ b/Demo/Sources/Model/URNMedia.swift
@@ -18,7 +18,7 @@ enum URNMedia {
     static let onDemandVerticalVideo = Media(
         title: "Vertical video",
         imageUrl: "https://www.rts.ch/2022/10/06/17/32/13444380.image/4x5",
-        type: .urn("urn:rts:video:13444390")
+        type: .urn("urn:rts:video:13950405")
     )
     static let onDemandVideo = Media(
         title: "A bon entendeur",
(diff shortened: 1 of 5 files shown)


Demo/Sources/Model/URNMedia.swift
```diff
@@ -18,7 +18,7 @@ enum URNMedia {
     static let onDemandVerticalVideo = Media(
         title: "Vertical video",
         imageUrl: "https://www.rts.ch/2022/10/06/17/32/13444380.image/4x5",
```
waliid:
```suggestion
        imageUrl: "https://img.rts.ch/medias/2023/image/w2e9c5-26920079.image/9x16",
```

waliid (approved):

waliid (approved):

PR #1582, opened 2026-09-18
Support URL-based asset downloading

### LINKED ISSUE #1499: Offline simple asset support

defagos:
## Description

This PR makes it possible to download URL-based items. SwiftData is used for local storage, as for URN-based content (see #1512).

## Design and implementation considerations

A `URLAssetLoader` and `URLAssetDownloadStore` are provided by the PillarboxPlayer package. When playing or downloading a URL, we usually do not know:

- Whether the URL is played as-is or if a resource loader delegate, respectively a content key session delegate, are required.
- Whether custom data is required locally, not only to decide the right kind of asset to build (e.g. content key session delegate with persisted keys), but also for potential local tracking purposes.

For these reasons, `URLAssetLoader` and `URLAssetDownloadStore` are parametrized using respective providers:

- `URLAssetLoaderProvider`, which defines custom data that needs to be handled, as well as how this custom data can be used to decide which asset must be played or downloaded.
- `URLAssetDownloadStoreProvider`, inheriting from `URLAssetLoaderProvider`, which defines how a local asset can be created from local persisted custom data.

This design allows progressive disclosure, i.e. a `URLAssetLoaderProvider` can be implemented for URL-based content playback and further extended to a `URLAssetLoaderProvider` in case persistence is required.

When playing a URL, the input uniquely identifying the download is `URLInput`, which contains the URL to be played as well as the associated metadata (since all metadata must be provided immediately).

A special `URLEmptyAssetProvider` has been introduced to provide a straightforward API when no custom data is required. Note that in this case only simple assets are supported, as there is no metadata available to support the construction of a resource loader delegate or content key session delegate.

### Remark

The `URLAssetLoader` and `URLAssetDownloadStore` we provide are used together. Should implementations need to download URLs with a different kind of storage, they should implement their own (trivial) URL loader as well as a companion store directly. For this reason neither `URLAssetLoader` nor `URLAssetDownloadStore` are provided as public APIs. Only their associated provider protocols are.

## Changes made

- Added `URLAssetLoader` and its associated `URLAssetLoaderProvider` protocol.
- Added `URLAssetDownloadStore ` and its associated `URLAssetDownloadStore Provider` protocol.
- Added `URLDownloader`, parametrized with `URLAssetDownloadStoreProvider`, to manage downloads.
- Added an `AssetLoader` optional method to customize the asset to be downloaded from a source, should this asset be different from the asset to be played online. Note that this method returns a publisher, should downloadable asset creation require an asynchronous work (e.g. retrieval of a token).
- Added `URLInput` to convey inputs associated with a URL-based item.
- Added `URLEmptyAssetProvider` to support URL-based asset loading and downloads without custom metadata.
- Added `EmptyCustomData` to express the absence of custom data (required for persistence, `Void` or `Never` cannot work here).
- Updated `PlayerMetadata` to be an alias for `AssetMetadata` bearing `EmptyCustomData` accordingly.
- Updated `PlayerItem` APIs to support URL-based loading either without custom data (simple items) or with custom data/asset loading using a provider. The `.custom` and `.encrypted` items previously available have been accordingly removed.
- Updated URN asset loading to fetch the token asynchronously before a download.
- Made Core Business assets public. This makes it possible to implement asset loaders that ultimately deliver an asset played with SRG SSR Akamai token protection or FairPlay encryption.
- Enabled downloads in nightly builds.
- Ensured demo download settings are not provided in tvOS builds.
- Updated demo media handling to account for the new APIs.
- Replaced URL download support previously implemented in the demo with the new APIs.
- Updated PR template.

## Breaking changes

- Custom and encrypted `PlayerItem`s cannot be created directly anymore. A provider must be used instead.

## Checklist


Sources/Player/Downloads/URLAssetDownloadStore.swift
```diff
@@ -0,0 +1,212 @@
...
+            self.metadata = .init(assetMetadata: record.input.metadata)
+            self.bookmarkData = record.bookmarkData
+            self.progress = record.progress
+            self.error = .init(error: record.error)
+            self.creationDate = record.creationDate
+        }
+    }
+}
+
+@_spi(DownloaderPrivate)
```
waliid:
```suggestion
```

waliid (approved):

